PR #1, opened 2026-04-16
feat: add passthrough_env config to whitelist env vars through scrub

KE7:
Adds a `passthrough_env` config option that preserves specified environment variables through HELIX env scrub for both evaluator commands and Claude Code mutation sessions.

**Problem:** `CUDA_VISIBLE_DEVICES` and other critical env vars were being stripped during env scrub, breaking GPU-dependent workloads.

**Solution:** New `passthrough_env` list in config that whitelists env vars to survive scrubbing.

### DIFF
--- a/examples/circle_packing/helix.toml
+++ b/examples/circle_packing/helix.toml
@@ -1,6 +1,10 @@
 [project]
 objective = "Maximize the sum of radii of 26 non-overlapping circles packed in a unit square. Higher total radius sum = better packing."
 
+# Uncomment to pass specific env vars through the scrub into evaluator
+# and Claude Code subprocesses (e.g. GPU selection, cache dirs):
+# passthrough_env = ["CUDA_VISIBLE_DEVICES", "MUJOCO_GL", "HF_HOME"]
+
 [evaluator]
 command = "python3 evaluate.py"
 score_parser = "json_score"
--- a/src/helix/config.py
+++ b/src/helix/config.py
@@ -304,6 +304,14 @@ class HelixConfig(BaseModel):
     objective: str
     seed: str = "."
     rng_seed: int = 0  # GEPA parity: deterministic RNG for selection
+    passthrough_env: list[str] = Field(
+        default_factory=list,
+        description=(
+            "Environment variable names to pass through the env scrub into "
+            "evaluator and Claude Code subprocesses (e.g. "
+            '["CUDA_VISIBLE_DEVICES", "MUJOCO_GL", "HF_HOME"]).'
+        ),
+    )
     evaluator: EvaluatorConfig
     dataset: DatasetConfig = Field(default_factory=DatasetConfig)
     seedless: SeedlessConfig = Field(default_factory=SeedlessConfig)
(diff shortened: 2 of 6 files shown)


KE7 (commented):
## Review: `passthrough_env` config — ✅ Looks good (would approve if not own PR)

Clean, well-scoped PR. The implementation is correct and the security model is sound. One DRY concern worth addressing in a follow-up.

### 1. Correctness ✅
All three call-sites pass `passthrough_env` through correctly:
- **Evaluator** (`executor.py`): `_scrub_environment(split, ..., passthrough_env=config.passthrough_env)`
- **Mutator** (`mutator.py:mutate`): `invoke_claude_code(..., passthrough_env=config.passthrough_env)`
- **Merger** (`merger.py:merge`): same pattern
- **Seed generation** (`mutator.py:generate_seed`): also covered

### 2. Security ✅
- **Opt-in by default**: empty list means zero behavioral change for existing configs.
- **Allowlist model**: only explicitly named vars pass through — no glob/prefix matching that could accidentally leak secrets.
- Config-file-only surface: an attacker who can modify `helix.toml` already has code exec, so this adds no new attack surface.

### 3. Backward Compatibility ✅
- `default_factory=list` means existing configs without `passthrough_env` get `[]`, which produces identical scrubbed envs. Test `test_passthrough_env_empty_list_no_change` confirms this.

### 4. Test Coverage ✅ (with caveat)
Three solid tests cover `_scrub_environment` passthrough:
- Listed vars are preserved
- Missing vars silently skipped
- Empty list == no change

**Caveat:** The *new* inline scrub in `invoke_claude_code` (mutator.py lines 354-368) has **no dedicated tests**. It duplicates the same PATH/HOME/HELIX_*/passthrough logic but isn't exercised by the executor security tests. Consider adding a test that verifies the CC subprocess receives the expected env.

### 5. Code Quality — one concern

**DRY violation in `invoke_claude_code`:** The new env-scrub block in `mutator.py` (lines 354-368) manually reimplements the same logic as `executor._scrub_environment`. If the scrub policy changes (e.g., adding `TMPDIR` or `LANG`), you'd need to update both places. Consider either:
- Extracting a shared `_base_scrub(passthrough_env)` helper (without HELIX_SPLIT/HELIX_INSTANCE_IDS) into a common module, or
- Calling `_scrub_environment("unused", passthrough_env=...)` and dropping the HELIX_SPLIT key after

Not blocking on this — it's a small duplication today — but worth a follow-up before the scrub logic grows.

**Minor style nit:** `import os` in `mutator.py` is placed after `logger = logging.getLogger(__name__)` and before the package imports, breaking PEP 8 grouping. Move it up to the stdlib import block at the top.

### 6. Edge Cases ✅
- **Missing env var**: silently skipped (`if key in os.environ`). Tested.
- **Duplicates in list**: harmless — last write wins with same value.
- **HELIX_* in passthrough**: no conflict — the passthrough loop runs after the HELIX_* loop, so it would just overwrite with the same value.

**Verdict: Approve.** Ship it, then consider extracting the shared scrub helper.

PR #3, opened 2026-04-17
feat: add StratifiedBatchSampler for task-diverse minibatches

KE7:
## Motivation

HELIX currently uses upstream reference `EpochShuffledBatchSampler`, which draws minibatches uniformly at random from the training set. When the training set contains multiple groups with many cases per group (e.g. `group_alpha__case_0`, `group_alpha__case_1`, ..., `group_beta__case_0`, ...), a minibatch of size K will often be dominated by a single group, especially with short batches and skewed group-size distributions.

For reflection-style evolutionary search, that's a problem: the LLM mutator reasons about whichever group the batch happens to contain, so single-group-dominated batches push evolution toward overfitting one group rather than cross-group generalisation.

## Design

`StratifiedBatchSampler` (new, HELIX-specific: **not** a upstream reference parity change) guarantees that every minibatch of size K contains instances from **K distinct groups**, where the group key is derived from each instance id via a user-supplied `group_fn`.

- **Epoch layout**: at the start of each epoch, shuffle instances within each group, then interleave across groups round-robin. The resulting flat schedule has the property that any contiguous minibatch-sized window aligned to a minibatch boundary touches K distinct groups.
- **Trimming (not padding)**: if a round would emit fewer than K distinct groups (only happens at the tail when `num_groups % K != 0`), those entries are dropped rather than padded; padding would duplicate a group and break stratification. At most `num_groups % K` instances are dropped per round.
- **Epoch bumps** reshuffle via the shared `rng`, preserving determinism parity with `EpochShuffledBatchSampler`.
- **Fallback**: when `len(groups) < minibatch_size`, stratification is impossible, so the sampler transparently delegates to an internal `EpochShuffledBatchSampler`; behaviour is identical to the current default.

### Config

Two new options under `[evolution]`:

- `batch_sampler`: `"epoch_shuffled"` (default, upstream reference parity) | `"stratified"`
- `group_key_separator`: default `"__"`, which splits instance ids on this separator and takes the first part as the group key (e.g. `group_alpha__case_3` -> `group_alpha`).

Existing configs are unchanged: `batch_sampler` defaults to `epoch_shuffled`, so there is no behavioural difference unless the user opts in.

### Wiring

`src/helix/evolution.py` now picks between `EpochShuffledBatchSampler` and `StratifiedBatchSampler` based on `config.evolution.batch_sampler`. Both samplers share the same `rng` so the rng-consumption order relative to `candidate_selector` remains unchanged on the default path (upstream reference parity preserved).

## Tests

All 466 existing tests still pass. 8 new tests in `tests/unit/test_batch_sampler.py`:

- [x] `test_stratified_minibatch_contains_k_distinct_groups`: every K=3 batch from `{a__0, a__1, b__0, b__1, c__0, c__1}` has all 3 groups.
- [x] `test_stratified_epoch_covers_all_instances`: balanced groups: every instance yielded once per epoch.
- [x] `test_stratified_epoch_rotation_reshuffles`: epoch bump triggers a fresh shuffle.
- [x] `test_stratified_fallback_when_groups_lt_minibatch_size`: falls back to `EpochShuffledBatchSampler` when only 2 groups but K=3.
- [x] `test_stratified_fallback_epoch_covers_all_instances`: fallback path still covers every instance.
- [x] `test_stratified_empty_loader_raises`: empty loader raises `ValueError` (parity with `EpochShuffledBatchSampler`).
- [x] `test_stratified_determinism_same_seed`: identical rng seeds produce identical minibatch sequences.
- [x] `test_stratified_unbalanced_group_sizes_still_stratifies`: unbalanced group sizes (3/2/2) still yield K-distinct-group minibatches.

## Test plan

- [x] `pytest tests/ -x`: all 466 tests pass locally
- [ ] Reviewer: confirm no upstream reference parity behaviour is altered on the default path

### DIFF
--- a/src/helix/batch_sampler.py
+++ b/src/helix/batch_sampler.py
@@ -115,9 +115,9 @@ class EpochShuffledBatchSampler(Generic[DataId]):
 class StratifiedBatchSampler(Generic[DataId]):
     """Task-stratified minibatch sampler.
 
-    Each minibatch of size ``K`` is constructed by choosing ``K`` distinct
-    groups (per ``group_fn``) and then picking one unused instance from each
-    of those groups.  This guarantees task diversity within every minibatch,
+    Each epoch is pre-shuffled such that every minibatch of size ``K`` is a
+    slice of ``K`` instances drawn from ``K`` distinct groups (per
+    ``group_fn``).  This guarantees task diversity within every minibatch,
     which helps reflection-style evolutionary search reason about
     generalisation instead of overfitting to whichever task happens to
     dominate a random batch.
@@ -130,14 +130,19 @@ class StratifiedBatchSampler(Generic[DataId]):
       any contiguous window of ``minibatch_size`` indices begins at a
       multiple of ``minibatch_size`` and touches ``minibatch_size`` distinct
       groups — as long as at least ``minibatch_size`` groups exist.
+    - Per-round rotation: when ``num_groups > minibatch_size``, each round
+      drops the trailing ``num_groups % minibatch_size`` groups (padding
+      would re-introduce a group collision).  We rotate ``group_keys`` by
+      ``r`` positions on round ``r`` so the dropped slot rotates across all
+      groups within a single epoch — preventing within-epoch group
+      starvation.
     - Epoch bumps reshuffle via the shared ``rng``, mirroring
       :class:`EpochShuffledBatchSampler`'s determinism guarantees.
-    - Padding: when the combined interleaved schedule is not a multiple of
-      ``minibatch_size``, we drop the trailing partial round rather than
-      pad with duplicates, because padding would re-introduce a group
-      collision within the final minibatch.  Because all full rounds are
-      emitted first, this only ever discards a tail of size
-      ``< minibatch_size`` per epoch.
+    - Padding: when a round is partial (some buckets exhausted), we trim to
+      the largest multiple of ``minibatch_size`` that still fits in the
+      round, dropping any trailing remainder rather than padding with
+      duplicates (padding would re-introduce a group collision within the
+      final minibatch).
     - Fallback: when ``len(groups) < minibatch_size``, a stratified minibatch
       is impossible, so the sampler transparently delegates to an internal
       :class:`EpochShuffledBatchSampler` for GEPA parity semantics.
@@ -145,9 +150,14 @@ class StratifiedBatchSampler(Generic[DataId]):
     Invariants:
       S1. Every returned minibatch of size ``m`` contains exactly ``m``
           distinct group keys (when ``num_groups >= m``).
-      S2. Across a full epoch, each instance is yielded at most once
-          (instances beyond ``floor(min_group_size * num_groups / m) * m``
-          may be dropped in the final partial round).
+      S2. Within an epoch each instance is yielded at most once.  In each
+          round of the round-robin interleave, only the first
+          ``whole_rounds_per_round = (num_groups // m) * m`` entries are
+          kept (rounded down to a multiple of ``m`` for partial rounds).
+          When ``num_groups % m != 0``, the trailing ``num_groups % m``
+          group slots in any given round are dropped — but the per-round
+          rotation of ``group_keys`` ensures the dropped slot rotates
+          across all groups, so no group is starved within an epoch.
       S3. Determinism given ``rng`` seed, matching
           :class:`EpochShuffledBatchSampler` semantics.
     """
@@ -199,34 +209,34 @@ class StratifiedBatchSampler(Generic[DataId]):
             self.rng.shuffle(buckets[key])
         self.rng.shuffle(group_keys)
 
-        # Interleave round-robin: round r picks buckets[k][r] for each key k,
-        # as long as index r is valid for that bucket.  A "full round" touches
-        # every group exactly once — that's our stratification guarantee.
+        # Interleave round-robin and trim per-round.  Round r picks
+        # buckets[k][r] for each key k where r is a valid index for that
+        # bucket.  A "full round" touches every group exactly once — that's
+        # our stratification guarantee.  Within a round, the first
+        # ``whole_rounds_per_round = (num_groups // m) * m`` entries form
+        # complete stratified minibatches; any trailing ``num_groups % m``
+        # entries are dropped because padding them would require
+        # duplicating a group.  When a round is partial (some buckets
+        # exhausted), we further clip to a multiple of ``m`` so the slice
+        # formula in ``next_minibatch_ids`` never overruns a boundary.
+        #
+        # Per-round rotation: rotating ``group_keys`` by ``r`` positions
+        # before slicing rotates which group lands in the trimmed slot
+        # across rounds, so no group is starved within an epoch when
+        # ``num_groups > m`` and ``num_groups % m != 0``.
         max_rounds = max(len(buckets[k]) for k in group_keys)
-        schedule: list[DataId] = []
-        for r in range(max_rounds):
-            round_slice: list[DataId] = []
-            for key in group_keys:
-                if r < len(buckets[key]):
-                    round_slice.append(buckets[key][r])
-            # Only emit the round if it fills at least one full minibatch of
-            # distinct groups; otherwise stratification would break.
-            if len(round_slice) >= self.minibatch_size:
-                schedule.extend(round_slice)
-
-        # Trim to a whole-number-of-minibatches length so the slice formula
-        # in ``next_minibatch_ids`` never overruns a boundary.  Each full
-        # round has length ``num_groups``; within a round, the first
-        # ``(num_groups // m) * m`` entries form complete stratified
-        # minibatches — any trailing ``num_groups % m`` entries are dropped
-        # because padding them would require duplicating a group.
         m = self.minibatch_size
         whole_rounds_per_round = (num_groups // m) * m
         trimmed: list[DataId] = []
-        # Split ``schedule`` back into rounds of size ``num_groups``.
-        for start in range(0, len(schedule), num_groups):
-            round_ids = schedule[start : start + num_groups]
-            trimmed.extend(round_ids[:whole_rounds_per_round])
+        for r in range(max_rounds):
+            offset = r % num_groups
+            rotated_keys = group_keys[offset:] + group_keys[:offset]
+            round_slice = [buckets[k][r] for k in rotated_keys if r < len(buckets[k])]
+            # Take at most whole_rounds_per_round entries, then clip to a
+            # multiple of m in case this round is partial.
+            keep = min(len(round_slice), whole_rounds_per_round)
+            keep -= keep % m
+            trimmed.extend(round_slice[:keep])
 
         self.shuffled_ids = trimmed
 
@@ -236,6 +246,13 @@ class StratifiedBatchSampler(Generic[DataId]):
         trainset_size = len(loader)
         if trainset_size == 0:
             raise ValueError("Cannot sample a minibatch from an empty loader.")
+
+        # Fast path: once the fallback is engaged and the trainset hasn't
+        # changed, delegate directly without re-bucketing all_ids() on every
+        # call.  ``self.epoch`` is left untouched in this branch — the inner
+        # fallback maintains its own epoch state.
+        if self._fallback is not None and trainset_size == self.last_trainset_size:
+            return self._fallback.next_minibatch_ids(loader, state)
 
         base_idx = state.i * self.minibatch_size
         curr_epoch = (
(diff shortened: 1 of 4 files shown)


KE7 (commented):
## Review: `feat: add StratifiedBatchSampler for task-diverse minibatches`

**Verdict: LGTM.** No blockers. A few non-blocking notes below.

### 1. Correctness — distinct-group guarantee ✅

Traced the round-robin interleave + trim logic by hand and with a local repro (5 groups of sizes (3,3,3,2,2), K=3): every minibatch yielded 3 distinct groups, across epoch boundaries as well. The key invariants hold:

- Within `_update_shuffled`, a round r emits `round_slice = [buckets[k][r] for k in group_keys if r < len(buckets[k])]`, which by construction contains distinct group keys. Partial rounds are only appended when `len(round_slice) >= minibatch_size`, so `round_ids[:whole_rounds_per_round]` always takes distinct-group entries.
- `whole_rounds_per_round = (num_groups // m) * m` correctly drops the `num_groups % m` tail per round — padding would duplicate a group, so trimming is the right call. Docstring explains this well.
- Final asserts in `next_minibatch_ids` (`len(shuffled_ids) % m == 0`, `end_idx <= len(shuffled_ids)`) are a nice safety net.

### 2. Edge cases — fewer groups than K ✅

Fallback to internal `EpochShuffledBatchSampler` is clean and lazily constructed. Empty-loader `ValueError` parity is preserved. Single-group / single-instance cases also land in the fallback branch correctly.

### 3. Epoch rotation & coverage ⚠️ (nit)

- For **balanced groups** with `num_groups == m`: full coverage per epoch, test confirms ✅.
- For **unbalanced groups** with `num_groups == m`: some tail instances dropped per epoch as documented in S2; within-group shuffle + shared rng means different instances get dropped each epoch, so long-run coverage is ergodic. Test `test_stratified_unbalanced_group_sizes_still_stratifies` verifies stratification but not that all instances are eventually seen.
- For **`num_groups > m`**: one group's entire round-r slot gets trimmed per round. Since `group_keys` is shuffled once per epoch, which group gets dropped rotates across epochs — confirmed by repro above. Not tested explicitly.

**Suggestion (non-blocking):** add one test that runs N epochs with `num_groups > m` and asserts every instance id appears at least once in the union of all emitted minibatches. This would document/lock in the "eventually covered" behaviour.

### 4. Config wiring ✅

`EvolutionConfig.batch_sampler` (Literal) + `group_key_separator` (with empty-string validation in `model_post_init`) are the right shape. Default `"epoch_shuffled"` preserves existing configs unchanged. Closure in `evolution.py` uses the `_sep=sep` default-arg idiom to capture the separator safely — good.

One tiny nit: no integration test verifies that setting `[evolution] batch_sampler = "stratified"` in a config actually routes to `StratifiedBatchSampler` in `run_evolution`. The unit tests cover the sampler in isolation; the wiring path is only exercised by manual inspection. Low priority.

### 5. Test coverage ✅

Eight tests cover: distinct-group guarantee, epoch coverage, epoch rotation, fallback path, fallback coverage, empty-loader `ValueError`, determinism, and unbalanced stratification. Reasonable. Gaps noted above.

### 6. Backward compatibility ✅

- `batch_sampler` defaults to `"epoch_shuffled"`, so unmodified configs produce byte-identical behaviour.
- In `evolution.py`, the type annotation changed from `EpochShuffledBatchSampler[str] | None` to `BatchSampler[str] | None` (the existing protocol). On the default path, the same `EpochShuffledBatchSampler` is instantiated with the same shared `rng`, preserving rng-consumption order relative to `candidate_selector` — upstream reference parity is intact.
- Note: the "stratified" path consumes `rng` differently (one `shuffle` per bucket + one `shuffle` of group_keys, vs a single `shuffle` of all ids). That's fine since parity is only claimed on the default path.

### 7. Minor observations (optional)

- **Fallback perf nit:** when `num_groups < minibatch_size`, `_update_shuffled` is invoked on every call because `shuffled_ids` stays empty → `needs_refresh` is always `True`. Each call re-buckets the full `all_ids()`. Not a correctness issue (no rng consumed in the fallback branch of `_update_shuffled`), but inelegant. Could early-return from `next_minibatch_ids` straight to `self._fallback.next_minibatch_ids(...)` once `_fallback is not None` and `last_trainset_size == trainset_size`.
- **`curr_epoch` in fallback mode:** `base_idx // max(len(shuffled_ids), 1) == base_idx` when `shuffled_ids` is empty. The outer `self.epoch` thus increments on every call. Harmless because the inner fallback has its own epoch state, but worth a code comment if you touch this again.
- **Determinism:** `self.rng.shuffle(group_keys)` mutates the returned-sorted list in place — fine because it's a local. Just flagging for future readers.

### CI

Both `test (3.11)` and `test (3.12)` green. 🎉

---

Nice work — clean, well-documented, conservative defaults. Not blocking the merge on any of the above.

KE7 (commented):
## Summary

Overall the design is sound and the motivation is compelling — `EpochShuffledBatchSampler`'s non-stratified behavior is a real problem for multi-task evolutionary search. Default path is unchanged (upstream reference parity preserved) and CI is green. However, I found **one correctness bug that crashes the sampler** on a realistic config (uneven task sizes with `num_groups > minibatch_size`), plus a within-epoch group-starvation issue and a few smaller items. The existing test set does not exercise the failing configuration because the only unbalanced test uses `num_groups == minibatch_size`.

CI status: ✅ `test (3.11)` and `test (3.12)` both pass.

---

## MUST-FIX

### 1. `_update_shuffled` trim logic mis-aligns when any included round has `m ≤ len(round_slice) < num_groups`

**File:** `src/helix/batch_sampler.py:202-231`

The round builder emits any round where `len(round_slice) >= self.minibatch_size`, so rounds can be shorter than `num_groups` (when some groups have already been exhausted). The trim loop then re-chunks the flat `schedule` in strides of `num_groups`:

```python
for start in range(0, len(schedule), num_groups):
    round_ids = schedule[start : start + num_groups]
    trimmed.extend(round_ids[:whole_rounds_per_round])
```

This is only correct if every emitted round has exactly `num_groups` entries. When a partial round of size `[m, num_groups)` gets emitted, the chunk boundaries no longer align with round boundaries — so the trim can (a) produce a `trimmed` length that is not a multiple of `minibatch_size`, tripping `assert len(self.shuffled_ids) % self.minibatch_size == 0` in `next_minibatch_ids` (L258), and (b) even without the assert, return minibatches that span two rounds and repeat a group.

**Reproducer** (fails on all seeds 0–9):

```python
import random
from helix.batch_sampler import StratifiedBatchSampler

class Loader:
    def __init__(self, ids): self.ids = ids
    def all_ids(self): return list(self.ids)
    def __len__(self): return len(self.ids)

class State:
    def __init__(self, i): self.i = i

ids = ['a__0','a__1','a__2','b__0','b__1','c__0','d__0','d__1','d__2']  # num_groups=4, m=2, uneven
loader = Loader(ids)
s = StratifiedBatchSampler(minibatch_size=2, group_fn=lambda x: x.split('__',1)[0],
                           rng=random.Random(0))
s.next_minibatch_ids(loader, State(0))   # AssertionError: len 9 not multiple of 2
```

With seed 0, `shuffled_ids = ['a__0','c__0','b__1','d__0','a__2','b__0','d__2','a__1','d__1']` — length 9, crashes.

This is a realistic config: HELIX's motivating example has many tasks with varying numbers of seeds. A dataset with e.g. `specific example id` (3 seeds), `pick_place` (2 seeds), `open_drawer` (1 seed), `stack_blocks` (3 seeds) at `minibatch_size=2` would crash on the first minibatch.

**Suggested fix:** iterate rounds with their actual lengths and take at most `whole_rounds_per_round` from each round directly, without going through a flat `schedule` list. For example:

```python
m = self.minibatch_size
whole_rounds_per_round = (num_groups // m) * m
trimmed: list[DataId] = []
for r in range(max_rounds):
    round_slice = [buckets[k][r] for k in group_keys if r < len(buckets[k])]
    # Only keep a prefix that fits whole minibatches of distinct groups.
    keep = min(len(round_slice), whole_rounds_per_round)
    keep -= keep % m
    trimmed.extend(round_slice[:keep])
self.shuffled_ids = trimmed
```

Also add a regression test that matches the reproducer above.

---

## SHOULD-FIX

### 2. Within-epoch group starvation when `num_groups > minibatch_size`

**File:** `src/helix/batch_sampler.py:200, 217-231`

`group_keys` is shuffled once per epoch and then `(num_groups // m) * m` entries are kept from each round. That means `num_groups % m` groups — whichever happen to land in the trailing shuffled positions — see **zero** instances for the entire epoch. Demonstration with `num_groups=5, m=3`:

```
shuffled_ids (epoch 0) = [d__1, c__1, e__0, d__2, c__2, e__1, d__0, c__0, e__2]
groups used: {c, d, e}  — groups 'a' and 'b' skipped entirely this epoch
```

Between epochs a new shuffle rotates which groups get dropped, so over enough epochs each group amortizes to `m/num_groups` presence. But for the short runs the PR motivates (reflection-style evolutionary search), this disparity can be meaningful. Docstring invariant S2 (L148-150) does not reflect this — it says only "instances beyond `floor(min_group_size * num_groups / m) * m` may be dropped in the final partial round", which both (a) is arithmetically off and (b) obscures that whole groups are dropped.

**Suggested fix:** rotate `group_keys` per round (e.g. `group_keys[r % num_groups:] + group_keys[:r % num_groups]`) or sample `m` distinct groups per round so every group gets a turn. Also tighten the S2 docstring to describe actual behavior.

### 3. Fallback-path `self.epoch` counter grows unboundedly; `_update_shuffled` called every step

**File:** `src/helix/batch_sampler.py:240-251`

When `num_groups < m`, `shuffled_ids` is left empty. Then:

```python
curr_epoch = base_idx // max(len(self.shuffled_ids), 1)  # = i*m when fallback
needs_refresh = not self.shuffled_ids or ...            # always True in fallback
```

So every call re-enters `_update_shuffled` (re-buckets all ids — O(n) `group_fn` calls per step), and `self.epoch` increments by `m` per call. Observed:

```
i=0 epoch=0
i=1 epoch=3
i=2 epoch=6
i=3 epoch=9
```

No external correctness impact (the inner `_fallback` tracks its own state), but it's wasteful on large trainsets and makes `sampler.epoch` meaningless for anyone inspecting it. Suggest: once fallback is engaged, short-circuit to `return self._fallback.next_minibatch_ids(loader, state)` at the top of `next_minibatch_ids` unless `trainset_size != self.last_trainset_size`.

### 4. Missing test for `num_groups > m` + unbalanced groups

**File:** `tests/unit/test_batch_sampler.py:240-251`

`test_stratified_unbalanced_group_sizes_still_stratifies` uses `num_groups == m == 3`, which does not exercise the trim logic failure in (1). Please add:

- A test with `num_groups > m` and uneven group sizes (the reproducer above) that asserts `len(shuffled_ids) % m == 0` and no minibatch has duplicate groups.
- A test with `num_groups > m` balanced groups that asserts every group is eventually visited across a small number of epochs (guards against #2 regressing further).
- Optional: an integration-level test that `config.evolution.batch_sampler = "stratified"` gets wired through `run_evolution` correctly (the config/wiring path is currently untested end-to-end).

---

## NITS

### 5. Class docstring wording — L118-120

> Each minibatch of size `K` is constructed by choosing `K` distinct groups (per `group_fn`) and then picking one unused instance from each of those groups.

Actually the schedule is pre-built once per epoch; nothing is "chosen" at `next_minibatch_ids` time. Minor, but misled me briefly. Suggest: "Each epoch is pre-shuffled such that every minibatch of size `K` is a slice of `K` instances drawn from `K` distinct groups."

### 6. `evolution.py:809-810` — default-arg capture is unnecessary

```python
def _group_fn(example_id: str, _sep: str = sep) -> str:
    return example_id.split(_sep, 1)[0]
```

Plain closure over `sep` is simpler and equivalent here (`sep` is not rebound in the surrounding scope).

### 7. `config.py:290-294` — `group_key_separator` empty-string guard runs even on `epoch_shuffled`

Harmless because the default is `"__"`, but conceptually the validation is only meaningful when `batch_sampler == "stratified"`. Move the check into the stratified branch, or at least clarify in the docstring.

### 8. `batch_sampler.py:149-150` — invariant S2 formula

`"instances beyond floor(min_group_size * num_groups / m) * m may be dropped"` — this expression does not match what the code actually drops (see #2). Please rewrite in terms of `whole_rounds_per_round = (num_groups // m) * m` and explicitly note whole-group drops when `num_groups % m != 0`.

---

## Positives

- Default path is untouched; no upstream reference parity regression on `batch_sampler="epoch_shuffled"`.
- Shared rng with `candidate_selector` is preserved on the default path — the comment at `evolution.py:791-800` shows care for the differential-testing harness.
- Fallback to `EpochShuffledBatchSampler` when `num_groups < m` is a good safety valve.
- Tests cover several useful properties (distinct-groups, determinism, empty-loader, fallback, epoch rotation).
- Good docstrings overall and clean TRACE integration.

Holding back approval until the crash bug and the missing test are addressed; happy to re-review after.

PR #8, opened 2026-04-19
Validate config strictly; treat example_ids as opaque strings in Architecture A

KE7:
## Summary

Two coherent fixes, both in the "don't silently corrupt user input" bucket:

1. **Reject unknown keys in config sub-models.** Add `extra='forbid'` to every pydantic sub-model in `src/helix/config.py` so misplaced or mistyped TOML keys raise `ValidationError` at load time instead of being silently dropped.
2. **Treat `example_ids` as opaque strings end-to-end in Architecture A.** Remove the `int()` cast at the `helix_batch.json` handoff boundary so structured ids like `"group_alpha__case_3"` (emitted by `StratifiedBatchSampler`) survive the round-trip.

## Fix 1: config strictness

In a real `helix.toml`, a user placed these two keys under `[evaluator]`:

```toml
[evaluator]
command = "..."
batch_sampler = "stratified"         # wrong section!
group_key_separator = "__"           # wrong section!
```

Both keys actually live on `EvolutionConfig` (see `src/helix/config.py:273-293`), not `EvaluatorConfig`. Because pydantic defaults to `extra='ignore'`, the validator silently dropped them and the run proceeded with the default `epoch_shuffled` sampler: exactly the opposite of what the user asked for, with **zero warning**.

The same class of silent failure hits typos (`batch_samplr`), stale fields from old configs, and section-name drift.

The fix: every sub-model now has `model_config = ConfigDict(extra="forbid")`: `EvaluatorConfig`, `DatasetConfig`, `SeedlessConfig`, `EvolutionConfig`, `ClaudeConfig`, `WorktreeConfig`, and top-level `HelixConfig`. Any undeclared key at any nesting level is surfaced through the existing `load_config` friendly-error path.

## Fix 2: opaque string ids in Architecture A

`_write_helix_batch` in `src/helix/evolution.py` previously cast every element through `int()` at the JSON serialization boundary:

```python
def _write_helix_batch(worktree_path, indices: list[int]) -> None:
    path.write_text(json.dumps([int(i) for i in indices]))
```

and both call sites pre-cast too:

```python
_write_helix_batch(candidate.worktree_path, [int(s) for s in example_ids])
```

This silently made `StratifiedBatchSampler` effectively unreachable on Architecture A. The sampler's whole point is to emit structured ids like `"group_alpha__case_3"` so each minibatch spans K distinct groups, and `int("group_alpha__case_3")` raises `ValueError` the instant those ids hit the handoff.

The fix passes ids through opaquely:

- `_write_helix_batch(worktree_path, example_ids: list[str])`: no cast.
- Both call sites drop their `[int(s) for s in ...]` comprehensions.
- Docstrings/comments shift from "positional-index handoff" to "example-id handoff".

## Breaking changes

The `helix_batch.json` payload shape changes from `list[int]` to `list[str]`. Evaluators that previously read it as:

```python
indices = json.loads(Path("helix_batch.json").read_text())
subset = dataset[indices]   # list[int] -> positional indexing
```

now need to cast on their side:

```python
ids = [int(s) for s in json.loads(...)]
```

or switch to string-keyed lookup. The default `_RangeDataLoader` still emits `"0"`, `"1"`, ... so runs using `dataset.train_size` continue to work after the evaluator-side cast.

Helix is pre-1.0 (`0.1.2`); both breakages (strict config + string-id payload) are acceptable at this stage. Users with cruft in their `helix.toml` will get an explicit error instead of silent ignore, and evaluator authors using Architecture A already had to read the handoff file explicitly: adding one `int()` at their end is a minimal, one-line change. Callers who don't use Architecture A are unaffected.

## Why `extra='forbid'` is the right strictness pre-1.0

The only "cost" is that users with cruft in their `helix.toml` files get an explicit error instead of silent ignore, which, for a config parser whose job is to faithfully represent user intent, is the feature, not a regression. The alternative (emit a warning) is strictly worse: warnings get scrolled past and the user still gets a run that doesn't match their config.

## Other changes

- Dropped a stale `parallel_eval = false` entry from a fixture in `tests/unit/test_config.py`: the field was removed from `EvolutionConfig` long ago and was only surviving because of the silent-drop behaviour this PR removes.
- Updated stale "int-convertible ids" comments in `tests/unit/test_evolution.py` and `tests/unit/test_align_iteration.py`: that restriction is gone now.
- Rebased onto `main` after PR #7 landed; no merge conflicts.

## New tests

- `tests/unit/test_config_errors.py::TestUnknownKeysRejected`: three tests for config strictness:
  - `test_misplaced_batch_sampler_under_evaluator_exits`: the exact bug that motivated fix 1
  - `test_typo_in_evolution_section_exits`: `batch_samplr` typo
  - `test_unknown_top_level_key_exits`: `HelixConfig` itself
- `tests/unit/test_evolution_minibatch.py::TestWriteHelixBatchStringIds`: two tests for the handoff:
  - `test_structured_string_ids_round_trip`: neutral structured ids survive verbatim
  - `test_stringified_int_ids_still_written_as_strings`: regression guard for the breaking payload shape

## Test plan

- [x] `uv run pytest tests/ -q`: 532 passed (+5 new tests vs the main baseline).
- [x] `uv run mypy --strict src/helix/`: clean.
- [x] Every previously-passing test still passes, confirming no legitimate sub-model was relying on extra keys and no in-tree evaluator was relying on `list[int]` shape.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `DatasetConfig.train_size` / `val_size` — cardinality-only fields that drive
   the minibatch sampler when the evaluator owns the dataset (Architecture A
-  positional-index handoff).  HELIX writes sampled integer indices to
+  example-id handoff).  HELIX writes sampled example ids to
   `{worktree}/helix_batch.json`; the evaluator filters its own dataset.
 - `SeedlessConfig` — new section carrying `enabled` plus the optional
   prompt-grounding `train_path` / `val_path` used during seed generation.
 - Evaluator-owned dataset integration wired into Architecture A with
   cardinality-only `train_size` / `val_size`.
+- `load_config` now emits a dedicated hint for pydantic `extra_forbidden`
+  validation errors, pointing users at likely typos or misplaced keys.
 
 ### Changed
 - **BREAKING**: `seedless` is now a section (`[seedless]` with `enabled = …`),
@@ -23,6 +25,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **BREAKING**: `dataset.train_path` / `dataset.val_path` have moved to
   `seedless.train_path` / `seedless.val_path`.  `[dataset]` is now
   cardinality-only (`train_size` / `val_size`).
+- **BREAKING**: `helix_batch.json` payload shape is now `list[str]` instead
+  of `list[int]`.  Example ids flow through the Architecture A evaluator
+  handoff as opaque strings — the default `_RangeDataLoader` emits `"0"`,
+  `"1"`, …, and `StratifiedBatchSampler` emits task-prefixed ids like
+  `"cube_stack__3"`.  Evaluators that previously read the handoff as
+  `list[int]` must cast on their side
+  (`[int(s) for s in json.loads(Path("helix_batch.json").read_text())]`)
+  or switch to string-keyed lookup.  Unblocks the stratified sampler on
+  Architecture A, which previously died with
+  `ValueError: invalid literal for int()` at the serialization boundary.
+- **BREAKING**: All pydantic sub-models in `src/helix/config.py`
+  (`EvaluatorConfig`, `DatasetConfig`, `SeedlessConfig`, `EvolutionConfig`,
+  `ClaudeConfig`, `WorktreeConfig`, `HelixConfig`) now use
+  `model_config = ConfigDict(extra="forbid")`.  Unknown / misplaced /
+  mistyped TOML keys raise a validation error at load time instead of
+  being silently dropped.  Previously, placing `batch_sampler` under
+  `[evaluator]` (the key lives on `[evolution]`) silently left users on
+  the default sampler with no warning.
 
 ## [0.1.0] - 2026-04-10
 
--- a/README.md
+++ b/README.md
@@ -250,8 +250,10 @@ protected_files = ["evaluate.py"] # optional extra files HELIX must keep immutab
 
 [dataset]
 # Cardinality of the train / val splits.  Used by HELIX's minibatch
-# sampler to generate integer indices that the evaluator (running in
-# the worktree) filters against its own dataset via helix_batch.json.
+# sampler to generate example ids (stringified indices by default, or
+# opaque "group__N" ids when evolution.batch_sampler = "stratified")
+# that the evaluator (running in the worktree) filters against its own
+# dataset via helix_batch.json — written as an opaque JSON list[str].
 # Leave both unset for single-task mode (legacy full-batch path).
 # train_size = 200
 # val_size  = 200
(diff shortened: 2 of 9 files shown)


KE7:
(Automated review by OMAR agent.)

## Overview
Two surgical fixes, well-motivated and complete:
1. `extra='forbid'` on every pydantic sub-model in `src/helix/config.py` (7/7 classes covered).
2. `example_ids` now flows through `_write_helix_batch` as opaque `list[str]`, unblocking `StratifiedBatchSampler`'s `"group__N"` ids on Architecture A.

CI green on 3.11 and 3.12; state MERGEABLE; the PR body clearly flags the on-disk `helix_batch.json` shape as a breaking change.

## High-severity issues
None. Both changes are correct, scoped, and reversible.

## Medium-severity issues

- **`README.md` documents the old behaviour.** Lines 253-254, 302, and 308 still call this "positional-index handoff" and say HELIX writes "integer indices" to `helix_batch.json`. Post-merge, the docs contradict the implementation. The `[dataset]` TOML comment block needs the same rewrite that was applied to `DatasetConfig`'s docstring.
- **`CHANGELOG.md` `[Unreleased]` does not mention this PR.** The on-disk payload shape change (`list[int]` to `list[str]`) and the `extra='forbid'` strictness change are exactly the class of thing downstream evaluator authors read changelogs for. Lines 11-14 even contain stale "positional-index" phrasing that is now inaccurate. Given the project is pre-1.0 the breakage is fine, but silent breakage through the changelog is the same failure mode this PR is fixing inside the config loader.

## Low-severity / nits

- **`load_config` has no tailored hint for `extra_forbidden` errors** (`src/helix/config.py:443-447`). Users hitting the exact footgun this PR targets (e.g. `batch_sampler` under `[evaluator]`) will see `Error: Extra inputs are not permitted` with no pointer to the correct section. A one-liner `elif error["type"] == "extra_forbidden": print("   Hint: key is not recognised on this section; check for typos or a misplaced sub-section")` would close the loop. Functional without it; the field path is already printed, which is why `test_misplaced_batch_sampler_under_evaluator_exits` passes.
- **Double-copy of id lists**: `list(example_ids)` at `evolution.py:696` and `:725`, then again inside `_write_helix_batch` via `json.dumps(list(example_ids))`. Negligible; one copy is enough.
- **`list(batch)` at `:725`** is fine but unnecessary; `batch` is already `list[str]` at the call site.

## What looks good

- Coverage of `extra='forbid'` is **complete**: all 7 `BaseModel` subclasses (`EvaluatorConfig`, `DatasetConfig`, `SeedlessConfig`, `EvolutionConfig`, `ClaudeConfig`, `WorktreeConfig`, `HelixConfig`) got the `model_config` line. No gaps.
- `TestWriteHelixBatchStringIds::test_structured_string_ids_round_trip` encodes the exact motivation with neutral structured ids; this is the test that would have caught the bug.
- `test_typo_in_evolution_section_exits` (`batch_samplr`) is a great regression guard against the typo-silent-drop class of failure.
- Docstring updates are thorough and internally consistent: `DatasetConfig`, `_RangeDataLoader`, `_write_helix_batch`, `_full_val_example_ids`, and the Architecture A comment block in `run_evolution` all now say "example-id handoff" / "opaque strings". No stale "positional-index" phrasing left in the source tree.
- Test-comment updates in `test_align_iteration.py:230` and `test_evolution.py:590` correctly reflect the new invariant (ids are opaque; `"1"`/`"2"` are just legible stand-ins).
- The existing `load_config` error path at `config.py:433-448` surfaces pydantic's `extra_forbidden` type without any additional wiring; good reuse.
- Dropping stale `parallel_eval = false` from `test_config.py` is a satisfying side-effect that proves the fixture was only passing because of the very silent-drop behaviour this PR removes.

## Overall recommendation

**Approve.** Correct, well-tested, pre-1.0 so the breaking payload shape is acceptable. Medium-severity findings are both **docs**, not code; I would ask the author to follow up with a README + CHANGELOG update (either in this PR or a quick follow-up), but neither is a merge blocker.

KE7:
(Automated follow-up by OMAR agent.)

Addressed the medium + low findings from the review above in 7c7b9af (pushed at reviewer-operator's direction, not the PR author's):

- **README.md** — rewrote the three stale "positional-index / integer indices" doc spots to match the new example-id / opaque-string handoff.
- **CHANGELOG.md `[Unreleased]`** — fixed stale phrasing and added two new BREAKING entries (`helix_batch.json` payload shape; `extra='forbid'` strictness across all pydantic sub-models) + an Added entry for the new `load_config` hint.
- **`src/helix/config.py`** — `load_config` now emits a dedicated hint for pydantic's `extra_forbidden` error type, pointing users at typos or misplaced sub-sections (the motivating `batch_sampler` under `[evaluator]` footgun now gets a targeted message instead of only `"Extra inputs are not permitted"`).
- **`src/helix/evolution.py`** — dropped three redundant `list(...)` copies in the `_write_helix_batch` path.
- **`tests/unit/test_config_errors.py`** — added a regression test for the new hint.

Validation: `uv run pytest tests/ -q` → **533 passed** (+1 new); `uv run mypy --strict src/helix/` → clean.

No PR title/body edits — those remain with the author.

PR #10, opened 2026-04-22
[codex] support multiple mutation backends

KE7:
## What changed
- replace the backend-specific mutation config with an `agent` config section and remove the backward-compat `claude` alias
- add backend support for Claude Code, Codex CLI, Cursor Agent, Gemini CLI, and OpenCode in the HELIX mutator/merge flow
- persist normalized backend artifacts so HELIX keeps session, tool-usage, token, and cost data across supported backends
- update docs/tests and bump the package version from `0.1.3` to `0.2.0`

## Why
HELIX previously only supported Claude Code. The goal of this change is to let HELIX launch the other installed backends in yolo/headless mode while preserving the output files needed for usage accounting.

## Impact
Users can now select `agent.backend` in `helix.toml` or pass `--backend` on the CLI. This is a breaking config change because the supported config section is now `[agent]` rather than `[claude]`.

## Validation
- `uv sync --all-extras`
- `uv run pytest tests/unit/ -q`
- `uv run mypy --strict src/helix/`
- manual smoke tests of the installed headless backend commands for codex, cursor, gemini, and opencode; claude command shape verified but the local account was rate-limited during execution

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 **Evolutionary optimization for full codebases using agentic coding tools as the mutation engine and git worktrees as the population pool.**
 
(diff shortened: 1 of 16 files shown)


KE7:
@codex review

PR #12, opened 2026-04-27
Add Docker sandbox runner support

KE7:
Following the Rogue Claude Code Sonnet agent incident, we are introducing docker sandboxing where there is one ephemeral container for the mutator, and one for the evaluator


KE7:
@claude @codex Review the PR.

PR #14, opened 2026-05-07
Fix upstream reference cache conformance divergences 29-34

KE7:
## Summary
- Default evaluation caching to off to match upstream reference Optimize Anything behavior.
- Key evaluation cache by candidate content rather than candidate id.
- Includes the tree-key refinement from former stacked PR #15: cache identity now uses `git rev-parse HEAD^{tree}` so reuse depends on tracked file content, not commit metadata/history.
- Make cache loading tolerant of corrupt/unexpected persisted cache data.
- Preserve actual uncached evaluation accounting and expose top-k best-example history for reflection diagnostics.

## Validation
- Focused cache/default/persistence/minibatch tests passed in branch audit.
- Budget tests passed in branch audit.
- Tree-key tests passed after including #15 changes.

### DIFF
--- a/src/helix/config.py
+++ b/src/helix/config.py
@@ -343,10 +343,10 @@ class EvolutionConfig(BaseModel):
         ),
     )
     cache_evaluation: bool = Field(
-        default=True,
+        default=False,
         description=(
-            "Enable (candidate_hash, example_id) evaluation cache. "
-            "GEPA parity: EngineConfig.cache_evaluation."
+            "Enable content-addressed evaluation caches. Defaults off, matching "
+            "GEPA Optimize Anything's conservative cache_evaluation behavior."
         ),
     )
     acceptance_criterion: Literal["strict_improvement", "improvement_or_equal"] = Field(
--- a/src/helix/evolution.py
+++ b/src/helix/evolution.py
@@ -625,22 +625,58 @@ def _worktree_lock(worktree_path: str | Path) -> threading.Lock:
 
 
 def _candidate_content_key(candidate: Candidate) -> str:
-    """Return a stable content key for cache reuse across equivalent candidates."""
+    """Return a stable content key for cache reuse across equivalent candidates.
+
+    Contract: the caller is responsible for ensuring the candidate's worktree
+    has its evaluation-relevant content committed (clean working tree, no
+    untracked files). The key is derived from ``HEAD^{tree}``, which is
+    content-addressable over the *committed* tracked tree only — it does not
+    reflect uncommitted modifications, the staged index, or untracked files.
+
+    Defensive behavior: if the worktree is detected to be dirty (modified
+    tracked files or untracked, non-ignored files present), we fall back to
+    ``candidate.id`` so that two candidates with identical HEAD trees but
+    differing dirty state cannot collide on the same cache key. Submodule
+    contents are summarized as gitlinks inside the parent tree, so changing
+    the submodule pointer changes the key but in-place edits inside an
+    uncommitted submodule do not — keep submodule state committed for
+    reliable reuse.
+    """
     try:
-        result = subprocess.run(
+        tree_proc = subprocess.run(
             ["git", "rev-parse", "HEAD^{tree}"],
             cwd=candidate.worktree_path,
             check=True,
             capture_output=True,
             text=True,
         )
-        sha = result.stdout.strip()
-        if sha:
-            return sha
-    except Exception:
+        sha = tree_proc.stdout.strip()
+        if not sha:
+            return candidate.id
+
+        # Reject the tree SHA if the worktree is dirty or has untracked
+        # (non-ignored) files; otherwise we'd risk a false cache hit.
+        status_proc = subprocess.run(
+            ["git", "status", "--porcelain=v1", "--untracked-files=normal"],
+            cwd=candidate.worktree_path,
+            check=True,
+            capture_output=True,
+            text=True,
+        )
+        if status_proc.stdout.strip():
+            logger.warning(
+                "Candidate %s worktree is not clean; falling back to id cache "
+                "key to avoid stale-content cache hits.",
+                candidate.id,
+            )
+            return candidate.id
+        return sha
+    except Exception as exc:
         logger.warning(
-            "Could not resolve git tree for candidate %s; falling back to id cache key.",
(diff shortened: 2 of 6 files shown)


KE7:
@claude review

KE7:
# Code Review — PR #14: Fix upstream reference cache conformance divergences 29-34

## Overview
This PR brings HELIX's evaluation cache closer to upstream reference / Optimize Anything semantics:
- Defaults `cache_evaluation` to `False`.
- Switches the per-example cache key from candidate `id` to a *content key* (git HEAD SHA of the candidate worktree, with a fallback to `id`).
- Gates the no-example/single-task `EvaluationCache` on the same `cache_evaluation` flag, so it can be `None`.
- Makes pickle-backed cache loading tolerant of corrupt/non-dict payloads (warn + return `None` rather than raise).
- Tightens budget accounting on seed re-evaluation (`was_cached=…`) and parent train re-evaluation.
- Adds `_inject_top_k_best_example_history(...)` to expose per-example top-K frontier scores via `per_example_side_info` for reflection prompts.
- Adds tests for cache-disabled, content-key reuse, and tolerant-load behavior.

Overall the direction looks right and the tests cover the key semantics. A few correctness and performance concerns are worth addressing before flipping out of draft.

## Correctness

- **Subprocess in the hot path / silent identity collisions (`_candidate_content_key`).** Every cache lookup now shells out to `git rev-parse HEAD` in `candidate.worktree_path`. Two issues:
  1. **Uncommitted worktree state is invisible.** If anything in the candidate worktree is modified but unstaged/uncommitted, the SHA stays the same and the cache will return *stale* results for genuinely different content. If candidates are guaranteed clean by construction, please assert that loudly (or add a quick `git status --porcelain` check / use a tree hash like `git rev-parse HEAD^{tree}` plus an index hash). Otherwise the new "content-addressed" cache can silently reuse the wrong evaluation.
  2. **Failure fallback mixes key spaces.** On subprocess failure we fall back to `candidate.id`. That means a single repo where `git` becomes briefly unavailable can pollute the cache with mixed key types (SHA vs id). Consider either failing closed (no cache write/read) or deriving the fallback from a stable hash of the worktree contents, not the lineage id.
  3. The `except Exception:` is very broad — at minimum log `exc` and prefer `(subprocess.CalledProcessError, OSError, FileNotFoundError)` so unexpected errors aren't swallowed.

- **`_inject_top_k_best_example_history` reaches into `ParetoFrontier._results`.** Private-attribute access from another module is fragile; please add a small public accessor (e.g. `frontier.iter_results()` returning `(cid, EvalResult)` pairs) and use it here and in the test guard. The current check `if … not frontier._results` will also throw `AttributeError` if the internal name is ever renamed.

- **`per_example_side_info` length re-init.** When `per_example != len(example_ids)`, we silently throw away the existing entries and rebuild from empty dicts. That can drop legitimate side-info on a length mismatch. At minimum log a warning, ideally raise — a length skew here is almost certainly a bug upstream.

- **Budget accounting symmetry.** `_eval_parent` previously did **not** add `_evaluation_budget_units()` for the cached parent train re-eval (the old code dropped the bool). The new code now charges budget when uncached, which is the right fix — but please double-check there is no other site (resume path, reflective_mutation) where the same train eval was already charged; otherwise some configs will see a one-time budget bump.

- **Behavioral default flip.** Changing `cache_evaluation` default from `True → False` is a user-visible semantics change. Worth calling out in the PR body / CHANGELOG / docs and double-checking integration tests / examples that may have relied on caching.

- **Tolerant cache load is good** — but consider also rotating the corrupt file aside (e.g. `eval_cache.pkl.corrupt-<ts>`) so users can post-mortem rather than have it silently overwritten on next save.

## Performance

- `git rev-parse` on every `_cached_eval` and every `_cached_evaluate_batch` call adds a fork+exec per candidate per split. For large minibatch loops this is non-trivial. Cache the SHA per `Candidate` (e.g. memoize on `candidate.id` in a module-level `WeakValueDictionary` or `functools.lru_cache`-friendly wrapper) and invalidate when a worktree is mutated.
- `_inject_top_k_best_example_history` is `O(|frontier| * |examples|)` per call; fine for the current scale but worth noting if the frontier grows.

## Style / Conventions

- `import subprocess` is added but the helper still uses `shutil`/`shlex` via the broader module — fine. Consider grouping the new helper near other git/worktree utilities for discoverability.
- `_inject_top_k_best_example_history` mutates and returns `eval_result`. Either return `None` (mutate-only) or return a new `EvalResult` (pure). The current "both" pattern is easy to misuse.
- Docstring on the new helper says "Optimize Anything exposes top-performing example histories" — citing the upstream reference/upstream reference file:line (as elsewhere in the file) would aid future audits.
- The cache-key dict literal `{"content_key": …, "split": …}` is constructed in two places (`_cached_evaluate_batch`, tests). Extracting a `_mb_cache_dict(candidate, split)` helper would prevent future drift between call sites and tests.

## Tests

- New coverage is good: cache-disabled path, content-key reuse, tolerant load (both non-dict and unreadable). Nice.
- Missing: a regression test that asserts the **fallback** path (`git rev-parse` fails → key is `candidate.id`) doesn't collide with a real SHA-keyed entry. Given the silent-collision risk above, this is worth pinning.
- Missing: a test asserting the seed-eval budget is charged exactly once when `cache_evaluation=True` and zero on a cache hit.
- The mocker patch `helix.evolution._candidate_content_key` is a good seam — consider exposing it as a real injection point on the cache rather than via monkeypatch in production callers.

## Security

- `subprocess.run([...], cwd=candidate.worktree_path)` uses a list (no shell=True), so no injection. `worktree_path` is trusted internal state — fine.
- Tolerant `pickle.load` still executes arbitrary pickle ops on whatever is at `eval_cache.pkl`; this was already true and is unchanged. If untrusted state directories are ever a concern, switching to a non-pickle format would be the long-term fix.

## Suggested Follow-ups (non-blocking)
1. Replace `git rev-parse HEAD` with `git rev-parse HEAD^{tree}` (or include index/dirty status) and memoize per candidate.
2. Add a `ParetoFrontier.iter_results()` (or similar) public accessor and remove `_results` access.
3. Document the `cache_evaluation` default flip in the PR description / release notes.
4. Quarantine corrupt `eval_cache.pkl` instead of silently dropping it.

LGTM on intent; please address the silent stale-cache risk and the private-attribute access before un-drafting.

PR #15, opened 2026-05-07
Key upstream reference evaluation cache by git tree

KE7:
## Summary
- Refines the upstream reference cache conformance branch by using git tree identity rather than commit HEAD identity for candidate content cache keys.
- Uses the tracked file tree hash so cache reuse ignores commit metadata/history but changes when tracked content changes.
- Keeps fallback behavior when git tree resolution is unavailable.

## Validation
- Focused TestCachedEvaluateBatch.
- Full tests/unit.
- ruff check src tests.
- git diff --check.

### DIFF
--- a/src/helix/evolution.py
+++ b/src/helix/evolution.py
@@ -625,22 +625,58 @@ def _worktree_lock(worktree_path: str | Path) -> threading.Lock:
 
 
 def _candidate_content_key(candidate: Candidate) -> str:
-    """Return a stable content key for cache reuse across equivalent candidates."""
+    """Return a stable content key for cache reuse across equivalent candidates.
+
+    Contract: the caller is responsible for ensuring the candidate's worktree
+    has its evaluation-relevant content committed (clean working tree, no
+    untracked files). The key is derived from ``HEAD^{tree}``, which is
+    content-addressable over the *committed* tracked tree only — it does not
+    reflect uncommitted modifications, the staged index, or untracked files.
+
+    Defensive behavior: if the worktree is detected to be dirty (modified
+    tracked files or untracked, non-ignored files present), we fall back to
+    ``candidate.id`` so that two candidates with identical HEAD trees but
+    differing dirty state cannot collide on the same cache key. Submodule
+    contents are summarized as gitlinks inside the parent tree, so changing
+    the submodule pointer changes the key but in-place edits inside an
+    uncommitted submodule do not — keep submodule state committed for
+    reliable reuse.
+    """
     try:
-        result = subprocess.run(
+        tree_proc = subprocess.run(
             ["git", "rev-parse", "HEAD^{tree}"],
             cwd=candidate.worktree_path,
             check=True,
             capture_output=True,
             text=True,
         )
-        sha = result.stdout.strip()
-        if sha:
-            return sha
-    except Exception:
+        sha = tree_proc.stdout.strip()
+        if not sha:
+            return candidate.id
+
+        # Reject the tree SHA if the worktree is dirty or has untracked
+        # (non-ignored) files; otherwise we'd risk a false cache hit.
+        status_proc = subprocess.run(
+            ["git", "status", "--porcelain=v1", "--untracked-files=normal"],
+            cwd=candidate.worktree_path,
+            check=True,
+            capture_output=True,
+            text=True,
+        )
+        if status_proc.stdout.strip():
+            logger.warning(
+                "Candidate %s worktree is not clean; falling back to id cache "
+                "key to avoid stale-content cache hits.",
+                candidate.id,
+            )
+            return candidate.id
+        return sha
+    except Exception as exc:
         logger.warning(
-            "Could not resolve git tree for candidate %s; falling back to id cache key.",
+            "Could not resolve git tree for candidate %s (%s); falling back "
+            "to id cache key.",
             candidate.id,
+            exc,
         )
     return candidate.id
 
--- a/src/helix/evolution.py
+++ b/src/helix/evolution.py
@@ -628,7 +628,7 @@ def _candidate_content_key(candidate: Candidate) -> str:
     """Return a stable content key for cache reuse across equivalent candidates."""
     try:
         result = subprocess.run(
-            ["git", "rev-parse", "HEAD"],
+            ["git", "rev-parse", "HEAD^{tree}"],
             cwd=candidate.worktree_path,
             check=True,
             capture_output=True,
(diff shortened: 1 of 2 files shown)


KE7:
@claude review

KE7:
## Code Review: Key upstream reference evaluation cache by git tree

### Overview
This PR refines the candidate content cache key in `_candidate_content_key` so it uses the git **tree** SHA (`HEAD^{tree}`) instead of the **commit** SHA (`HEAD`). This is a small, focused, and well-motivated change: tree identity is content-addressable, while commit identity also encodes parent, author, timestamp, and message — none of which affect evaluation results. Two new tests demonstrate the desired equivalence (empty/metadata-only commit → cache hit) and divergence (tracked content change → cache miss). Fallback to `candidate.id` on git failure is preserved.

### Strengths
- **Correctness of intent.** A git tree object is exactly the content-addressable identity we want for evaluation reuse: identical tracked-file trees ⇒ identical SHAs, regardless of commit metadata. This is the right primitive.
- **Minimal blast radius.** Only the argument to `git rev-parse` and the log message changed. Behavior is otherwise byte-identical, including the `Exception` fallback and the `len == 40` guard.
- **Strong test coverage of the semantics change.**
  - `test_tree_key_reuses_cache_across_different_commits_with_same_tree` directly exercises the previously-broken case (empty commit invalidates cache under `HEAD`, but not under `HEAD^{tree}`) and asserts both the call count and the cached scores.
  - `test_tree_key_misses_cache_when_tracked_content_changes` confirms the converse — content edits force re-evaluation.
  - Tests use a real on-disk git repo with `tmp_path`, which is far more convincing than mocking `subprocess.run`.
- **Helper extraction** (`_git`, `_init_repo`) keeps the new tests readable and is reusable for any future git-aware tests.
- **Log message updated** to match new semantics (\"git tree\" rather than \"git HEAD\"), avoiding future grep-confusion.

### Issues / Risks

**1. Uncommitted working-tree changes are still invisible to the cache key (medium).**
`HEAD^{tree}` reflects the *last commit's* tree, not the working tree or the index. If the candidate's worktree has staged or unstaged modifications relative to HEAD, two candidates with identical HEAD trees but different dirty states will collide on the same key, returning stale cached scores. This is the same limitation the original `HEAD` key had, so the PR doesn't regress, but it's worth documenting or hardening:
- Cheap mitigation: also incorporate `git status --porcelain=v1 -z` (or `git diff HEAD --raw`) hashed into the key, or refuse to cache when the worktree is dirty.
- Stronger mitigation: `git write-tree` (after `git add -A` in a scratch index) gives a true content tree of the current working state.
- At minimum, please add a one-line comment near `_candidate_content_key` noting the assumption (\"caller is responsible for ensuring the worktree is clean / committed before evaluation\").

**2. Untracked files are also not reflected.**
Even with a clean index, `HEAD^{tree}` ignores untracked files. If evaluation can depend on files that are typically untracked (data, generated artifacts, config), the cache key will under-discriminate. Worth confirming this matches the project's evaluation contract.

**3. Submodules / nested worktrees.**
If `worktree_path` ever points inside a submodule or a non-canonical worktree, `HEAD^{tree}` resolves the *outer* tree only when the cwd is correct; submodule contents are summarized as gitlinks (commit pointers) inside the parent tree. So changing a submodule's commit *does* change the parent tree (good), but changing files inside the submodule without a submodule commit does not. Probably fine for HELIX, but worth a sanity check.

**4. Minor: silent fallback severity.**
The `except Exception` branch logs at WARNING and falls back to `candidate.id`, which means *each candidate is its own cache key* — i.e., no cache reuse at all. In a long run, this could quietly tank evaluation throughput. Consider:
- Logging the underlying exception (`logger.warning(\"...: %s\", candidate.id, exc_info=True)` or `%s\", candidate.id, exc`), so debugging doesn't require reproducing.
- Possibly elevating to ERROR if this is unexpected in production.

**5. Test nit: unused locals.**
In `test_tree_key_reuses_cache_across_different_commits_with_same_tree`, `commit_a`, `commit_b`, `tree_a`, `tree_b` are computed and only `commit_a != commit_b` and `tree_a == tree_b` are asserted — this is fine and actually documents intent well, but ruff may flag the bindings if they ever change. Also consider adding an `assert _candidate_content_key(cand_a) == _candidate_content_key(cand_b)` to assert the key directly (in addition to the cache-hit behavior), which pins the contract independent of `_cached_evaluate_batch`.

**6. Test nit: helper signature.**
`_init_repo(path)` calls `path.mkdir()` unconditionally; if the directory already exists (e.g., a future test parameterization), it'll raise. Consider `path.mkdir(exist_ok=True)`. Low priority.

### Performance
Negligible. `git rev-parse HEAD^{tree}` is the same single subprocess call as `git rev-parse HEAD`. The cache benefit, however, is meaningful: empty/admin commits and rebases that preserve tree content will now hit the cache instead of triggering full evaluation.

### Security
No new attack surface. Inputs to subprocess are still a fixed argv with `cwd=candidate.worktree_path`; no shell, no user-controlled args.

### Conventions
- Matches surrounding code style: `subprocess.run([...], check=True, capture_output=True, text=True)`, `try/except Exception` with a logged warning and a typed fallback.
- Tests live in the same module/class as related cache tests and follow the same `mocker.patch` pattern.

### Recommendation
Approve once the **dirty-worktree assumption** is either documented in a code comment near `_candidate_content_key` or mitigated. The core change is correct, well-tested, and clearly an improvement over keying by commit SHA. Items (3)–(6) are nits and can be deferred.

KE7:
### Follow-up: review fixes pushed (fa7707c)

Addressed the items from the review on cdc8336:

**Source (`src/helix/evolution.py`)**
- Documented the contract of `_candidate_content_key` in its docstring (committed tracked content only; submodule semantics noted).
- Hardened the dirty-worktree case: after resolving `HEAD^{tree}`, run `git status --porcelain=v1 --untracked-files=normal`; if non-empty, log a warning and fall back to `candidate.id`. This prevents two candidates with identical HEAD trees but differing dirty / untracked state from colliding on the same cache key.
- Fallback warning now includes the underlying exception (`%s, exc`) so failures are debuggable without reproducing.

**Tests (`tests/unit/test_evolution_minibatch.py`)**
- `_init_repo` uses `mkdir(exist_ok=True)`.
- Added a direct `assert _candidate_content_key(cand_a) == _candidate_content_key(cand_b)` in `test_tree_key_reuses_cache_across_different_commits_with_same_tree` to pin the key contract independently of `_cached_evaluate_batch`.
- New test `test_content_key_falls_back_when_worktree_is_dirty` covers three states: clean (keys by tree SHA), modified-but-uncommitted tracked file (falls back to id), untracked file present (falls back to id).

**Validation**
- `ruff check src tests` → clean.
- Focused: `pytest tests/unit/test_evolution_minibatch.py -k "tree_key or content_key or CachedEvaluateBatch"` → 12 passed.
- Full: `pytest tests/unit` → 690 passed.

PR #16, opened 2026-05-07
Centralize Helix budget accounting

KE7:
## Summary
- Adds a centralized budget/progress accounting API in src/helix/budget.py.
- Routes evaluation, token/cost, generation, proposal, mutation, merge, and discovery counters through the shared API.
- Extends trace events with BUDGET_UPDATE metadata.
- Preserves current max-generation/evaluation behavior while making upstream reference-style budget hooks testable.

## Validation
- Full pytest passed in worker verification: 688 passed.
- mypy passed.
- changed-file ruff passed.

### DIFF
--- a/src/helix/budget.py
+++ b/src/helix/budget.py
@@ -1,16 +1,55 @@
-"""Central budget and progress accounting helpers for HELIX evolution."""
+"""Central budget and progress accounting helpers for HELIX evolution.
+
+Threading model
+---------------
+None of these helpers lock ``state``; callers serialize access by calling
+them only from the main evolution thread.  Today this invariant is
+maintained by ``evolution.py``: every ``budget.*`` call site is in the
+main loop, the presample/iteration loop, or a ``concurrent.futures.
+as_completed`` consumer (which runs serially on the submitting thread).
+The functions submitted to the parallel ``ThreadPoolExecutor``s
+(``_eval_parent``, ``_do_mutate``) are state-free — they neither read
+nor mutate ``EvolutionState``.
+
+If you ever move a ``budget.*`` call into a worker function body,
+introduce a ``threading.Lock`` here first; otherwise concurrent
+``next_mutation_id`` / ``next_merge_id`` calls can produce duplicate
+``g{gen}-s{n}`` ids and concurrent ``+=`` updates can drop charges.
+"""
 
 from __future__ import annotations
 
-from typing import Any
+import warnings
+from collections.abc import Mapping
+from typing import Any, TypedDict
 
 from helix.config import HelixConfig
 from helix.state import EvolutionState
 from helix.trace import TRACE, EventType
 
 
+class UsageDict(TypedDict, total=False):
+    """Subset of backend usage-payload fields read by ``charge_llm_usage``.
+
+    Pinning the keys makes the contract explicit at type-check time
+    without requiring callers to convert their existing
+    ``Candidate.usage: dict[str, Any]`` payloads.  ``charge_llm_usage``
+    accepts any read-only mapping so this remains backward-compatible.
+    """
+
+    input_tokens: int
+    output_tokens: int
+    cost_usd: float
+
+
 def budget_exhausted(state: EvolutionState, config: HelixConfig) -> bool:
-    """Return True if the configured evaluation budget is exhausted."""
+    """Return True iff the configured evaluation budget cap is set and reached.
+
+    A non-positive ``config.evolution.max_evaluations`` (``<= 0``)
+    disables the cap entirely; HELIX then runs until ``max_generations``
+    alone and this predicate always returns ``False``.  Mirrors GEPA's
+    metric-call cap (``optimize_anything.py`` ``max_metric_calls``).
+    """
     cap = config.evolution.max_evaluations
     return cap > 0 and state.budget.evaluations >= cap
 
@@ -18,7 +57,12 @@ def budget_exhausted(state: EvolutionState, config: HelixConfig) -> bool:
 def evaluation_budget_units(
     *, num_actual_examples: int | None = None, was_cached: bool = False
 ) -> int:
-    """Return evaluation budget units for an evaluation attempt."""
+    """Return evaluation budget units for an evaluation attempt.
+
+    Cached results charge 0 units; minibatch evals with N uncached
+    examples charge N (clamped at 0); single-task / no-example paths
+    charge 1 (one evaluator invocation).
+    """
     if was_cached:
         return 0
     if num_actual_examples is not None:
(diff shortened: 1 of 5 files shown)


KE7:
@claude review

KE7:
## Code Review: Centralize Helix budget accounting (#16)

### Overview
This PR introduces a new `src/helix/budget.py` module that centralizes all budget/progress accounting (evaluation count, token/cost rollups, generation index, monotonic proposal counter `state.i`, mutation/merge id allocation, discovery-budget stamping, and merge-invocation count). All scattered `state.budget.* +=` and counter-bumping sites in `evolution.py` are replaced with calls into the new API, and every charge emits a new `BUDGET_UPDATE` trace event. A new `tests/unit/test_budget.py` covers the API and uses a regex guard on `evolution.py` to keep raw counter mutations from leaking back in.

This is a good refactor — it removes a large category of "did we forget to bump the counter at this branch?" bugs and makes the upstream reference-style budget hooks first-class and testable. Behavior parity looks intentionally preserved.

### Strengths
- **Single source of truth.** Every counter mutation now goes through one named, documented function. The previous code had ~10 manual `state.budget.input_tokens += int(...)` triples that are easy to miss when adding a code path; the new API eliminates that footgun.
- **Tracing.** `BUDGET_UPDATE` events with `decision=<source>` give a full audit trail of why each unit was charged (`seed_val`, `merge_subsample`, `mutation_minibatch_gate`, `parallel_proposal`, etc.). Very nice for postmortems and upstream reference verification.
- **Behavior parity preserved.** `evaluation_budget_units` is identical to the old private helper (cached → 0, examples → max(0, n), else 1) and `budget_exhausted` keeps the `cap > 0` short-circuit so `max_evaluations <= 0` continues to disable the cap. The unit tests pin both.
- **Guardrail test.** `test_evolution_counter_mutations_route_through_budget_api` regex-asserts that the forbidden `state.budget.X +=`, `state.i +=`, etc. patterns never reappear in `evolution.py`. Cheap and high-value.
- **Discovery-budget stamping** (`record_discovery_budget`) is now used uniformly for seed/mutation/merge accept paths, eliminating three slightly-different inline assignments.

### Issues / Suggestions

**1. Thread-safety regression risk in parallel mutation path.**
`next_mutation_id`, `advance_proposal_counter` (called with `source="parallel_proposal"`), and `charge_llm_usage` are now invoked from what looks like a parallel proposal/mutation loop (`_p_idx > 0` branch and the futures section in `_do_mutate`). The old code had the same `state.X += 1` race (so this isn't *new*), but if the parallel path uses threads, two concurrent `next_mutation_id` calls can produce duplicate `g{gen}-s{n}` ids. Consider:
- Adding a `threading.Lock` inside `budget.py` around the counter bumps; or
- Documenting in the module docstring that callers must serialize counter-allocating calls; or
- Confirming the parallel executor is process-based (in which case state isn't shared) and noting that.

**2. `Event` dataclass is now wide.** `trace.py`'s `Event` gained 13 new optional fields, all used only by `BUDGET_UPDATE`. As more event types accumulate fields this scales poorly. Consider a `metadata: dict[str, Any] | None` field, or per-event-type subclasses / a tagged union, before the next event type is added.

**3. `decision` field is overloaded as "source label".** `Event` already has a `source: str | None` field (reserved per its comment for `"file:line"` from `inspect`). The new code passes the human source label as `decision=source`. That's fine for now but the naming is confusing — a future reader will assume `decision` is a per-iteration accept/reject decision. Either rename the parameter on the helpers (e.g., `reason` → mapped to a new `Event.reason` field) or add a comment to `Event` explaining the dual use.

**4. Two thin shims left in `evolution.py`.** `budget_exhausted` and `_evaluation_budget_units` in `evolution.py` are now one-line passthroughs to the API. Either:
- Inline the call sites and drop the local wrappers (cleanest), or
- Keep them but mark them `# re-export for backwards compat` so it's obvious they're not doing real work.

The local docstring on `budget_exhausted` is more informative than the API one — consider moving that docstring to `budget.budget_exhausted`.

**5. Trace volume.** Every charge — including `units == 0` (cached) and `usage == None` early-returns — now adds (or in the `None` case, doesn't add) a trace event. On a long run with frequent caching this can multiply trace size by ~5–10x compared to before. Two small mitigations to consider:
- Skip the `BUDGET_UPDATE` emit when `units == 0` (no-op charges); the tests don't depend on it.
- Or expose a `TRACE.set_level(...)` toggle so production runs can suppress per-charge events.

**6. `charge_llm_usage` early return is silent.** Returning on `not usage` is correct, but the previous code only entered the block under `if child.usage:` anyway. No bug, but the function signature now invites callers to pass `None` unconditionally and skip the gating. That's actually a feature; just worth a one-line note in the docstring ("safe to call with `None`; emits no event").

**7. `start_generation` mixes naming with `advance_proposal_counter`.** `start_generation` is a setter (`state.generation = generation`) while `advance_proposal_counter` is an incrementer. Consider `set_generation` for symmetry, or `begin_generation` if you want to imply lifecycle.

**8. `record_discovery_budget` overwrite semantics.** `state.num_metric_calls_by_discovery[candidate_id] = state.budget.evaluations` silently overwrites if called twice for the same id. The old inline code had the same behavior, so this is parity, but the helper is the right place to add an `assert candidate_id not in state.num_metric_calls_by_discovery` (or a debug warning) since duplicate inserts would indicate a real bug.

**9. Test gaps.**
- `test_progress_counters_update_through_budget_api` calls `advance_proposal_counter` once and asserts `state.i == 0`, which works because `EvolutionState.i` defaults to `-1` (upstream reference parity). A second call asserting `state.i == 1` would lock in the increment semantics explicitly.
- No test covers `charge_llm_usage` with `usage=None` (silent no-op) or with all-zero deltas (still emits an event today — is that intended?).
- No test covers `budget_exhausted` with a `was_cached=True` charge that adds 0 units near the cap (off-by-one regression risk).

**10. Type-hint nit.** `charge_llm_usage(usage: dict[str, Any] | None)` accepts an arbitrary dict. The repo presumably has a `UsageStats` TypedDict or dataclass already (referenced as `UsageStats()` in `evolution.py`); typing against that would be safer than `dict[str, Any]`.

### Risks
- **Parity:** Low. All arithmetic is preserved by `evaluation_budget_units`, and the regex guard test prevents drift.
- **Concurrency:** Pre-existing, but more visible now that all counters live in one module — flagging so a follow-up can make it explicit.
- **Trace size / log shape:** New event type may break downstream consumers that strict-validate event JSON shape (the new optional fields all default to `None`, so most consumers will be fine).
- **Draft status.** PR is marked draft; full pytest + mypy + changed-file ruff are reported green. Repo-wide ruff is blocked by an unrelated pre-existing F841.

### Verdict
Approve direction with minor follow-ups. The two things I'd most like to see before merge:
1. A clear answer on parallel-mutation thread-safety (lock vs. process-based vs. documented requirement).
2. Either inline or annotate the two `evolution.py` shims so the call surface is unambiguous.

Everything else is non-blocking polish.

PR #17, opened 2026-05-07
Test backend LLM budget usage accounting

KE7:
## Summary
- Adds test-only coverage proving backend usage payloads flow into EvolutionState.budget through charge_llm_usage.
- Covers claude, codex, cursor, gemini, and opencode parsing paths through mocked real backend outputs.
- Verifies seedless generation, mutation, and merge call sites charge centralized budget accounting.

## Validation
- Focused backend/budget tests: 17 passed.
- Affected suite: 154 passed.
- Full suite: uv run pytest -q => 696 passed.
- Separate installed-command smoke check confirmed all five backend CLIs are installed; true no-network live usage-output E2E is unavailable because live mutation paths invoke models.

### DIFF
--- a/tests/unit/test_budget.py
+++ b/tests/unit/test_budget.py
@@ -2,19 +2,38 @@
 
 from __future__ import annotations
 
-import re
 import json
+import re
+import subprocess
 from pathlib import Path
-from unittest.mock import MagicMock
 
 import pytest
 
-from helix.backends import BACKENDS
 from helix import budget
-from helix.config import AgentConfig, EvolutionConfig, EvaluatorConfig, HelixConfig
+from helix.backends import BACKENDS
+from helix.config import (
+    AgentConfig,
+    EvaluatorConfig,
+    EvolutionConfig,
+    HelixConfig,
+    SandboxConfig,
+)
 from helix.mutator import invoke_claude_code
 from helix.state import BudgetState, EvolutionState
 from helix.trace import TRACE, EventType
+
+
+# Mapping from backend name to the executable that ``_build_backend_args``
+# emits as args[0].  Used by parser-coverage tests to assert that
+# ``invoke_claude_code`` actually dispatched to the requested backend CLI
+# (rather than short-circuiting via the differential-testing override).
+_BACKEND_EXECUTABLE = {
+    "claude": "claude",
+    "codex": "codex",
+    "cursor": "cursor",
+    "gemini": "gemini",
+    "opencode": "opencode",
+}
 
 
 def make_state(evaluations: int = 0) -> EvolutionState:
@@ -76,7 +95,7 @@ def test_charge_evaluation_updates_counter_and_emits_event() -> None:
 @pytest.mark.parametrize(
     ("backend", "stdout", "expected"),
     [
-        (
+        pytest.param(
             "claude",
             json.dumps(
                 {
@@ -87,8 +106,9 @@ def test_charge_evaluation_updates_counter_and_emits_event() -> None:
                 }
             ),
             (11, 7, 0.31),
+            id="claude",
         ),
-        (
+        pytest.param(
             "codex",
             "\n".join(
                 [
@@ -100,8 +120,9 @@ def test_charge_evaluation_updates_counter_and_emits_event() -> None:
                 ]
             ),
             (12, 8, 0.32),
+            id="codex",
         ),
-        (
+        pytest.param(
             "cursor",
             "\n".join(
                 [
@@ -113,8 +134,9 @@ def test_charge_evaluation_updates_counter_and_emits_event() -> None:
                 ]
             ),
             (13, 9, 0.33),
+            id="cursor",
         ),
-        (
+        pytest.param(
             "gemini",
             "\n".join(
                 [
@@ -127,8 +149,9 @@ def test_charge_evaluation_updates_counter_and_emits_event() -> None:
                 ]
             ),
             (14, 10, 0.34),
+            id="gemini",
         ),
-        (
+        pytest.param(
             "opencode",
             "\n".join(
                 [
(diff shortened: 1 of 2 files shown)


KE7:
@claude review

KE7:
## Code Review — PR #17 (Test backend LLM budget usage accounting)

### Overview
This PR is purely additive test coverage (no production changes). It adds:
1. A parametrized unit test (`test_backend_usage_parsing_charges_llm_budget`) that drives `invoke_claude_code` with mocked `subprocess.run` stdout for all five supported backends (claude, codex, cursor, gemini, opencode) and verifies that `charge_llm_usage` updates `EvolutionState.budget` correctly.
2. Three integration tests in `TestLlmUsageBudgetIntegration` that confirm seedless generation, mutation, and merge call sites all route token/cost accounting through the central `budget_api.charge_llm_usage` API with the right `candidate_id` / `source` tags.

The intent is sound: lock in the contract between backend output parsers, `_normalise_usage_stats`, and the centralized budget API so future refactors cannot silently drop usage charges.

### Strengths
- **Realistic parser coverage.** Mocking at the `subprocess.run` boundary keeps the real `_parse_backend_output` + `_normalise_usage_stats` pipeline in the loop, so the tests catch alias-map regressions (e.g. `prompt_tokens` vs `inputTokens` vs `input`). Each backend's stdout shape mirrors the actual CLI format (top-level JSON for claude/opencode, JSONL for codex/cursor/gemini, advisory preamble + JSON for gemini).
- **Distinct numeric fixtures per backend** (11/7/0.31, 12/8/0.32, …) make a cross-wired alias regression easy to spot in failure output.
- **`mocker.spy`/`wraps=` on `charge_llm_usage`** is the right pattern: the original side effects still run, and `call_args_list` lets assertions verify both the `source`/`candidate_id` taxonomy and the resulting budget state. This is much stronger than asserting the budget delta alone.
- **PR description is honest** about why a true no-network E2E is impractical and lists the suite-level pass counts (17/154/696).

### Issues & suggestions

**Correctness / robustness**
- `mock_run.return_value = MagicMock(stdout=stdout, stderr=\"\", returncode=0)` will satisfy any attribute access. Prefer the real type so a future refactor that reads, e.g., `result.args` doesn't silently get a `MagicMock`:
  ```python
  mock_run.return_value = subprocess.CompletedProcess(
      args=[], returncode=0, stdout=stdout, stderr=\"\",
  )
  ```
- The parametrized test never asserts that `mock_run` was actually invoked with the expected backend executable. A trivial `assert mock_run.call_args.args[0][0] == backend_cli_name` (or `mocker.patch` of `_build_backend_args`) would guard against a regression where `invoke_claude_code` short-circuits via `_MUTATOR_OVERRIDE` and never reaches the parser at all. Today the test would still pass even if `_MUTATOR_OVERRIDE` were accidentally left set globally.
- In the three integration tests, `next(call for call in spy.call_args_list if call.kwargs.get(\"source\") == \"…\")` happily returns the first match. Consider asserting exactly-once for the candidate of interest, e.g.:
  ```python
  matches = [c for c in spy.call_args_list
             if c.kwargs.get(\"source\") == \"seed_generation\"
             and c.kwargs.get(\"candidate_id\") == \"g0-s0\"]
  assert len(matches) == 1
  ```
  Otherwise a future double-charge bug that *also* zeroes the second call's payload would slip through.
- The seedless test mocks `generate_seed` to return `usage` directly (`all_mocks[\"generate_seed\"].return_value = usage`). That bypasses `invoke_claude_code` entirely and only exercises `evolution.py`'s call site, which means *parser* regressions in seedless aren't covered. Either explicitly note this in a comment, or add one variant that lets `generate_seed` actually flow through `invoke_claude_code` with mocked subprocess output to assert end-to-end coverage of the seedless path.

**Coverage gaps worth a follow-up**
- `_normalise_usage_stats` also extracts `cached_input_tokens` and `reasoning_tokens`, but `charge_llm_usage` only sums `input_tokens`/`output_tokens`/`cost_usd`. If the budget API ever grows to count cached or reasoning tokens, this contract test is the natural place to capture it — a TODO-style xfail or a comment pointing at the omission would help.
- No assertion that the `BUDGET_UPDATE` trace event is emitted with the correct `decision`/`candidate_id` payload. `charge_llm_usage` already emits that event; spying on `TRACE.emit` (or checking the in-memory trace ring) would cement the observability contract.
- The sandbox path (`run_sandboxed_command`) is not covered by these tests. Worth adding at least one parametrize value that flips `sandbox=SandboxConfig(enabled=True)` and mocks `run_sandboxed_command` instead of `subprocess.run`.
- No negative test for `usage = {}` / `usage = None` — `charge_llm_usage` short-circuits in that case, but a regression that turns the early return into a `TypeError` would only surface in production.

**Style / project conventions**
- Import ordering in `tests/unit/test_budget.py`: `import json` should sort before `import re` (PEP 8 / isort default). Minor, but the rest of the file appears to be ruff/isort-clean.
- `from unittest.mock import MagicMock` is fine, but the rest of the test relies on `pytest-mock`'s `mocker`. Using `mocker.MagicMock()` keeps the dependency surface uniform.
- The parametrize id defaults to a stringified tuple of all three params, which produces noisy node ids. `pytest.param(..., id=\"claude\")` per case would make `pytest -k` invocations and CI logs much easier to read.
- Test class `TestLlmUsageBudgetIntegration` has no docstring; the rest of `test_evolution.py` documents class intent. A one-liner would help future readers.

**Performance / risk**
- Pure unit tests; negligible runtime impact. No security implications. No production code touched, so blast radius is limited to CI signal quality.
- Branch is targeted at `fix/upstream-budget-api`, not `main` — make sure that base lands first so this PR's diff stays focused on the test additions.

### Verdict
Solid, well-scoped test PR that meaningfully tightens the contract around centralized LLM usage accounting. Recommend addressing the **exactly-once spy assertion**, the **typed `CompletedProcess` mock**, and the **seedless parser coverage gap** before merge; the rest are nits / follow-ups.

— Generated via `/review`

PR #18, opened 2026-05-07
Guard incompatible upstream reference resume semantics

KE7:
## Summary
- Adds persisted resume_semantics metadata for optimization-meaningful config fields.
- Rejects resume when frontier type or other semantic settings would reinterpret saved state.
- Clears persisted eval cache when cache_evaluation is disabled so stale cache state is not revived later.
- Preserves Helix worktrees/directories as the artifact model.

## Validation
- Full pytest passed in worker verification: 687 passed.
- Focused resume/state/seedless tests: 34 passed.
- ruff check src tests passed.

### DIFF
--- a/examples/web_researcher/evaluate.py
+++ b/examples/web_researcher/evaluate.py
@@ -245,7 +245,7 @@ def evaluate():
             got = agent.solve(question_text)
             if not isinstance(got, str):
                 got = str(got) if got is not None else ""
-        except Exception as ex:
+        except Exception:
             got = ""
 
         # Score
--- a/src/helix/cli.py
+++ b/src/helix/cli.py
@@ -26,7 +26,7 @@ from helix.display import (
     print_warning,
     render_frontier_table,
 )
-from helix.exceptions import RateLimitError
+from helix.exceptions import RateLimitError, ResumeIncompatibleError, print_helix_error
 from helix.lineage import load_lineage
 from helix.population import EvalResult, FrontierType, ParetoFrontier, Candidate
 from helix.state import load_state, save_state
@@ -673,6 +673,13 @@ def evolve(
     setup_file_logging(base_dir)
     try:
         run_evolution(config, project_root, base_dir)
+    except ResumeIncompatibleError as exc:
+        # The current config would reinterpret persisted state.  Show the
+        # full Rich panel (with operation/phase/suggestion) instead of
+        # letting a raw traceback escape from the CLI.
+        logger.error("Resume rejected: %s", exc)
+        print_helix_error(exc)
+        raise SystemExit(2)
     except RateLimitError as exc:
         # Rate limits exhausted retries and bubbled all the way to the CLI.
         # State has been saved during the run; tell the user how to resume.
@@ -952,6 +959,18 @@ def resume(config_path: str, project_dir: Path | None) -> None:
     print_info(f"Resuming from generation {state.generation if state else 0}…")
     try:
         run_evolution(config, project_root, base_dir)
+    except ResumeIncompatibleError as exc:
+        logger.error("Resume rejected: %s", exc)
+        print_helix_error(exc)
+        raise SystemExit(2)
+    except RateLimitError as exc:
+        logger.error("Rate limit reached during resume: %s", exc)
+        print_error(
+            f"Rate limit reached: {exc}\n"
+            "Evolution state has been saved. "
+            "Run [cyan]helix resume[/cyan] again when rate limits clear."
+        )
+        raise SystemExit(2)
     except KeyboardInterrupt:
         _handle_keyboard_interrupt(project_root)
     else:
--- a/src/helix/evolution.py
+++ b/src/helix/evolution.py
@@ -64,6 +64,7 @@ from helix.population import (
 from helix.sandbox import start_evaluator_sidecar
 from helix.state import (
     BudgetState,
+    clear_eval_cache,
     EvaluationCache,
     EvolutionState,
     load_eval_cache,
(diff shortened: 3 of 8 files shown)


KE7:
@claude review

KE7:
## Code Review — PR #18: Guard incompatible upstream reference resume semantics

### Overview
The PR persists a small JSON-native `resume_semantics` snapshot inside `EvolutionState` and refuses to resume when the current `HelixConfig` would reinterpret saved state. It also clears the on-disk eval cache pickle when `cache_evaluation` is disabled so a later run can't revive stale cache entries. `state.py` schema bumps from v1 → v2; legacy payloads default-fill to `{}` and short-circuit validation. The companion test suite covers frontier-type rejection, semantic-config rejection, the cache toggle exemption, and `clear_eval_cache` removal.

Overall the design is sound and conservative: hard-reject is the right default for fields that change the meaning of persisted scores/frontier. Comments and parity references with upstream reference are very good. A few correctness-adjacent and UX issues below.

---

### Findings

**1. Eval cache is wiped before resume validation can refuse the run** (correctness / data loss)
In `evolution.py` the cache is loaded/cleared near line 1097–1105, *before* `load_state` + `_validate_resume_semantics` at 1123/1170:
```python
minibatch_cache = MinibatchEvalCache(...) if config.evolution.cache_evaluation else None
if minibatch_cache is None:
    clear_eval_cache(project_root)  # <-- destructive, runs unconditionally

# ... later ...
state = load_state(project_root)
_validate_resume_semantics(state, config)  # <-- may raise
```
If a user starts a resume with `cache_evaluation=False` *and* a semantic mismatch (e.g. wrong `frontier_type`), we delete `eval_cache.pkl` and *then* raise. When they correct the config and re-enable the cache, the persisted cache is gone. Suggest moving `clear_eval_cache` until after `_validate_resume_semantics` (i.e., after we've decided we're really proceeding with this config).

**2. ValueError surfaces as a raw traceback in the CLI** (UX)
`helix evolve` (cli.py:674–687) and `helix resume` (cli.py:953–958) only catch `RateLimitError` and `KeyboardInterrupt`. The new `ValueError("Cannot resume because…")` from `_validate_resume_semantics` will print a Python traceback rather than the friendly `print_error` users see for other failure modes. Consider either:
- Catching `ValueError` in both CLI sites and routing through `print_error` + `SystemExit(2)` (matching the rate-limit branch), or
- Defining a dedicated `ResumeIncompatibleError` and catching it explicitly so unrelated ValueErrors aren't swallowed.

**3. Behaviour change vs. previous `config_hash` warning** (compatibility)
Previously a config-hash mismatch only `print_warning(...)`'d and continued. Now the same edits trip a hard `ValueError` for any field in `resume_semantics`. This is intentional and correct, but it's a user-visible breaking change on resume semantics — worth a CHANGELOG / release-note entry and possibly a one-time migration tip in the error string (e.g. \"if you intentionally edited config, run `helix clean` and start fresh\").

**4. Redundant `clear_eval_cache` calls inside `_save_state`** (minor)
When caching is disabled, every `_save_state(s)` call invokes `clear_eval_cache(project_root)` — i.e., an `unlink()` of a path we already removed at startup. Harmless (FileNotFoundError swallowed) but unnecessary and slightly noisy on hot save paths. Could guard with a one-shot flag or just rely on the startup clear.

**5. `_semantic_diffs` doesn't recurse into lists** (minor / messaging quality)
The diff helper recurses into dict values but not list values, so list-shaped fields produce noisy `saved=[…long…], current=[…long…]` previews. Not incorrect — equality is still detected — but truncating or per-element diffing would make rejection messages friendlier, particularly for `evaluator.extra_commands` and `evaluator.protected_files`.

**6. Legacy state silently bypasses semantic validation** (acceptable, but document)
The early-return when `not state.resume_semantics` (evolution.py:192) is the right migration choice for v0/v1 states, but it means legacy resumes can still silently reinterpret state — until the next save back-fills `resume_semantics`. Suggest a one-line `print_info` (\"No persisted resume semantics found; this run's settings will be pinned for future resumes\") so the user knows the guard is dormant on this resume.

**7. State.json bloat** (minor)
`dataset.model_dump(mode=\"json\")` and the per-call rewrite of `resume_semantics` is fine in practice but does grow `state.json`. Worth a sanity-check on a representative project that the file stays well under any size assumptions logging/CI tooling makes.

**8. Field selection is conservative — confirm intent for two cases**
- `rng_seed`: changing it post-hoc is genuinely surprising, but mid-run it doesn't reinterpret persisted scores (RNG state isn't persisted yet). Conservative inclusion is fine; just confirm this is the intended trade-off vs. \"warn-only\".
- `merge_subsample_size` / `merge_val_overlap_floor`: these only affect *future* merge proposals; persisted scores remain meaningful. Hard-rejecting is defensible (saved decisions used the old values) but is the strictest of the available choices — make sure that's the policy you want.

**9. Tests** (coverage)
Coverage is good, but consider adding:
- Legacy state (`resume_semantics={}`) flowing through `_validate_resume_semantics` without raising (covers the early-return branch).
- An end-to-end CLI test of `helix resume` producing the friendly error path (ties in with finding #2).
- An ordering test that asserts `eval_cache.pkl` survives a rejected resume (ties in with finding #1).

---

### Smaller / nits
- `EvolutionState.resume_semantics` uses `dict[str, Any]`; no explicit `TypedDict`. Given this is the rejection contract, a `TypedDict` (or a small dataclass) would help future maintainers see what's authoritative without re-reading `_resume_semantics`.
- The PR description notes the full repo `ruff` is blocked by a pre-existing `F841` in `examples/web_researcher/evaluate.py`. Fine for this PR; consider opening a follow-up issue so that lint can run unconditionally in CI.
- Schema-version comment in `state.py` says \"subsequent bumps mark explicit JSON-native schema additions\" — confirm SCHEMA_VERSION=2 is also bumped if/when `resume_semantics` *contents* change, not only when fields are added/removed at the top level.

---

### Verdict
Good targeted fix with the right \"hard-fail by default\" stance for resume safety. The two items I'd treat as blocking are **#1 (cache wiped before validation can refuse the run)** and **#2 (raw traceback in CLI on rejection)** — both small to address. The rest are quality-of-life improvements.

KE7:
Addressed the review feedback in `aeab9fc` (pushed to `fix/upstream-resume-guards`).

### What changed
**Blocking fixes**
- **#1 cache wipe ordering**: removed the eager `clear_eval_cache` at startup; the disabled-cache cleanup now happens inside the first `_save_state` call (after `_validate_resume_semantics` has accepted the run). Regression test confirms `eval_cache.pkl` survives a rejected resume.
- **#2 CLI traceback**: introduced `ResumeIncompatibleError(HelixError)` with `operation`/`phase`/`suggestion` fields and added handlers in both `helix evolve` and `helix resume` that route through `print_helix_error` + `SystemExit(2)`. Also added a symmetric `RateLimitError` handler to `helix resume` (it was missing on that path).

**Quality of life**
- **#4** one-shot guard for the disabled-cache `clear_eval_cache` in `_save_state` (no more no-op `unlink()` on every save after the first).
- **#5** long-value truncation in `_semantic_diffs` previews so rejection messages don't get drowned by `evaluator.extra_commands` / `protected_files` dumps.
- **#6** `print_info` when resuming a legacy state with empty `resume_semantics`, so users know the guard is dormant for this resume but will be active afterward.
- Documented `_resume_semantics` as the authoritative resume-compat contract (top-level docstring lists each key).
- Friendlier suggestion in the rejection message (`helix clean` mentioned explicitly).

**Housekeeping**
- Fixed the `F841` in `examples/web_researcher/evaluate.py` — repo-wide `ruff check` now passes (it was previously called out in the PR description as a CI lint blocker).

### Validation
- `pytest`: **690 passed** (was 687; +3 new tests for legacy-state short-circuit, cache-survival on rejected resume, friendly CLI error path).
- `ruff check src tests`: clean.
- `ruff check .` (repo-wide): clean.

### Findings deliberately left as-is
- **#3 CHANGELOG entry**: no `CHANGELOG.md` exists in the repo today; I baked the user-visible warning into the rejection-message suggestion instead. Happy to add a `CHANGELOG.md` if you'd prefer.
- **#7 state.json bloat**: not measured here; flagging for a future check.
- **#8 policy questions** (`rng_seed`, `merge_subsample_size`): kept the conservative hard-reject behavior as written.
- **TypedDict for `resume_semantics`**: skipped to avoid invasive changes; the new docstring documents the shape instead.

KE7:
## Second-pass review after rebase onto main

Rebased `fix/upstream-resume-guards` onto `82db100` (main). One conflict to resolve in `src/helix/state.py`:
- **schema-version comment**: trivial wording overlap; combined the two notes into a single comment that documents the v0 migration path.
- **`clear_eval_cache` vs. `_quarantine_corrupt_cache`**: main added a quarantine helper (PR #14/#22) for corrupt eval-cache pickles; this branch added `clear_eval_cache` for the disabled-cache-toggle path. Both are needed and live next to each other now.

Auto-merged cleanly: `evolution.py`, `test_evolution_seedless.py`, `test_resume.py`, `test_state_persistence.py`.

After the rebase I re-audited `_resume_semantics` against the parity surface that landed in main (PR #14 tree-keyed cache + default-off `cache_evaluation`, PR #20 frontier acceptance, PR #16 budget API) and found three evaluator-side fields the original snapshot was missing.

### Parity gaps closed (`c4a8a08`)

| Field | Why it's resume-semantic |
|---|---|
| `evaluator.include_stdout` | Toggling changes which bytes reach the score parser, silently reinterpreting saved scores. |
| `evaluator.include_stderr` | Same as above. |
| `evaluator.sidecar` | Describes a completely different evaluator runtime (image, runner image, endpoint). Resuming under a swapped sidecar compares scores produced by different binaries. |

upstream reference gets strict equality on these implicitly by pickling its full config alongside state; HELIX has to enumerate them in `_resume_semantics`. The `_resume_semantics` docstring now also documents the **deliberately-omitted** surface so future maintainers can audit the contract at a glance:

- `evolution.cache_evaluation` — cache state lives in `eval_cache.pkl`.
- `evolution.max_generations` / `max_evaluations` / parallelism — budgets/concurrency, not score interpretation.
- `agent.*` (`backend`, `model`, `effort`, `max_turns`, `allowed_tools`, `background`) — only future proposals are affected; saved frontier scores stay valid (matches upstream reference treatment of the LM client as out-of-band).
- `sandbox` / `worktree` / top-level `seed` / `passthrough_env` / `env` — runtime/filesystem layout. Evaluator-environment drift is covered by the separate `evaluator_integrity_manifest`.

### New tests
- `test_resume_rejects_evaluator_stdout_capture_change`
- `test_resume_rejects_evaluator_sidecar_change`
- `test_resume_allows_agent_backend_change` — affirmative parity assertion that mutation-backend swaps are intentionally outside the contract.

### Validation
- `pytest`: **769 passed** (was 766 on rebased base; +3 new parity tests).
- `ruff check .` (repo-wide): clean.

### Branch state
- 3 commits on `fix/upstream-resume-guards` ahead of main:
  - `53f4af7` Guard incompatible resume semantics (rebased)
  - `592cf40` Address review feedback on resume-semantics guard (rebased)
  - `c4a8a08` Cover evaluator stdout/stderr capture and sidecar in resume guard (new)
- Force-pushed with `--force-with-lease`.

### Items left as policy decisions
- `rng_seed` and `merge_subsample_size` are kept hard-rejected (conservative, consistent with upstream reference's strict-equality stance).
- No `CHANGELOG.md` exists in the repo; the user-visible behavior change is communicated through the rejection-message suggestion.

PR #19, opened 2026-05-07
Return structured HelixResult from evolution

KE7:
## Summary
- Adds CandidateSummary and HelixResult programmatic result types.
- Changes run_evolution to return structured metadata instead of only the best Candidate.
- Exposes best candidate/result, candidate summaries, parents, scores, fronts, budgets, discovery counts, run dir, seed, and config hash.
- Keeps a compatibility id alias for callers that previously used best.id.

## Validation
- ruff, mypy, targeted evolution pytest, and full pytest passed in worker verification.
- Full pytest result reported by worker: 684 passed.
- Unrelated local worktree edits were preserved unstaged.

### DIFF
--- a/src/helix/__init__.py
+++ b/src/helix/__init__.py
@@ -2,7 +2,22 @@
 
 from importlib.metadata import PackageNotFoundError, version as _pkg_version
 
+from helix.population import (
+    Candidate,
+    CandidateSummary,
+    EvalResult,
+    HelixResult,
+)
+
 try:
     __version__ = _pkg_version("helix-evo")
 except PackageNotFoundError:  # not installed (e.g. source checkout without pip install -e .)
     __version__ = "0.0.0+unknown"
+
+__all__ = [
+    "Candidate",
+    "CandidateSummary",
+    "EvalResult",
+    "HelixResult",
+    "__version__",
+]
--- a/src/helix/evolution.py
+++ b/src/helix/evolution.py
@@ -872,15 +872,33 @@ def _build_helix_result(
     config: HelixConfig,
     lineage_path: Path,
 ) -> HelixResult:
-    """Build the structured programmatic result for a completed run."""
+    """Build the structured programmatic result for a completed run.
+
+    Reads the lineage JSON once at end-of-run.  ``record_entry`` flushes
+    every accept-time mutation/merge to disk synchronously, so the file
+    is the authoritative source by the time we get here.  Falls back to
+    each candidate's in-memory ``parent_ids`` / ``operation`` /
+    ``generation`` when no lineage entry is present (defensive — every
+    real accept site writes one).
+
+    Per-candidate data is materialized once on each
+    :class:`CandidateSummary`; :class:`HelixResult` exposes
+    ``aggregate_scores`` / ``sum_scores`` / ``instance_scores`` /
+    ``objective_scores`` / ``parents`` as cached views over that list,
+    so there is a single source of truth.
+    """
     lineage = load_lineage(lineage_path)
     non_dominated_ids = sorted(frontier.get_non_dominated())
     non_dominated = set(non_dominated_ids)
-    parents: dict[str, list[str]] = {}
     summaries: list[CandidateSummary] = []
 
-    for cid, candidate in frontier._candidates.items():
+    for cid, candidate in frontier.candidates.items():
         entry = lineage.get(cid)
+        # Parent resolution priority:
+        #   1. ``LineageEntry.parents`` — multi-parent (merge) source of truth.
+        #   2. ``Candidate.parent_ids`` — populated at construction for both
+        #      mutations and merges, used when lineage has nothing.
+        #   3. ``LineageEntry.parent`` — legacy single-parent fallback.
         candidate_parents = (
             list(entry.parents)
             if entry is not None and entry.parents
@@ -888,9 +906,8 @@ def _build_helix_result(
         )
         if not candidate_parents and entry is not None and entry.parent is not None:
             candidate_parents = [entry.parent]
-        parents[cid] = candidate_parents
 
-        result = frontier._results.get(cid)
+        result = frontier.get_result(cid)
         aggregate_score = result.aggregate_score() if result is not None else 0.0
         sum_score = result.sum_score() if result is not None else 0.0
         summaries.append(
@@ -918,25 +935,12 @@ def _build_helix_result(
     summaries.sort(key=lambda s: (s.generation, s.id))
     return HelixResult(
         best_candidate=best,
-        best_result=frontier._results.get(best.id),
-        candidates=list(frontier._candidates.values()),
+        best_result=frontier.get_result(best.id),
+        candidates=list(frontier.candidates.values()),
         candidate_summaries=summaries,
-        parents=parents,
-        aggregate_scores={
-            cid: result.aggregate_score() for cid, result in frontier._results.items()
-        },
-        sum_scores={cid: result.sum_score() for cid, result in frontier._results.items()},
-        instance_scores={
-            cid: dict(result.instance_scores) for cid, result in frontier._results.items()
-        },
-        objective_scores={
-            cid: (
-                list(result.objective_scores)
-                if result.objective_scores is not None
-                else None
-            )
-            for cid, result in frontier._results.items()
-        },
+        # Insertion-order copy of state.frontier — preserves accept order
+        # so callers can replay ``frontier_ids`` in chronological order.
+        # ``non_dominated_ids`` is sorted (above) for stable diffing.
         frontier_ids=list(state.frontier),
         non_dominated_ids=non_dominated_ids,
         frontier_type=state.frontier_type,
--- a/src/helix/evolution.py
+++ b/src/helix/evolution.py
@@ -51,7 +51,13 @@ from helix.executor import run_evaluator
 from helix.lineage import LineageEntry, find_merge_triplet, load_lineage, record_entry
 from helix.merger import merge, select_eval_subsample_for_merged_program
 from helix.mutator import mutate, build_seed_generation_prompt, generate_seed
-from helix.population import Candidate, EvalResult, ParetoFrontier
+from helix.population import (
+    Candidate,
+    CandidateSummary,
+    EvalResult,
+    HelixResult,
+    ParetoFrontier,
+)
 from helix.sandbox import start_evaluator_sidecar
 from helix.state import (
     BudgetState,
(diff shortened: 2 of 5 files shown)


KE7:
@claude review

KE7:
## Code Review — PR #19: Return structured HelixResult from evolution

### Overview
Replaces `run_evolution`'s `Candidate` return value with a new `HelixResult` dataclass plus `CandidateSummary`, exposing best candidate, per-candidate summaries, lineage, scores, frontier axes, budgets, run dir, seed, and config hash. A `_build_helix_result` helper assembles the structure inside `_run_evolution_impl`. Tests cover the happy path and the unrelated-but-needed `frontier_type="instance"` plumbing in `make_config`.

This is a clean upstream reference-parity API shape and all 684 tests reportedly pass. Below are concrete issues and suggestions.

---

### Correctness & API surface

- **Backwards-compat alias is partial.** `HelixResult.id` covers `best.id`, but any external caller that previously did `best.worktree_path`, `best.branch_name`, `best.generation`, `best.usage`, etc. on the return value of `run_evolution` will silently break (or raise `AttributeError`). The PR description only mentions `best.id`. Either:
  - expose `best_candidate` cleanly (which you already do — good), and call this out as a breaking change in CHANGELOG/docs, or
  - delegate a few more `Candidate` attributes via `__getattr__`/properties for a smoother migration, or
  - if you want strictness, leave it and note the break loudly.
  In-repo callers (`cli.py:675`, `cli.py:954`, and the seedless/minibatch/align tests) all discard the return value, so the in-tree blast radius is zero. The risk is downstream/notebook callers.

- **Public API export is missing.** `HelixResult`/`CandidateSummary` are now part of the documented return contract but I don't see them re-exported from `helix/__init__.py`. Worth adding so users can `from helix import HelixResult` rather than reaching into `helix.population`.

- **`run_evolution` is `@trace`-decorated style with two layers (`run_evolution` → `_run_evolution_impl`). Both signatures were updated to `-> HelixResult`. Good — consistent.

---

### Code quality

- **Reaching into `ParetoFrontier`'s private state.** `_build_helix_result` reads `frontier._candidates` and `frontier._results` (both leading-underscore). `ParetoFrontier` already exposes `get_non_dominated()`, `__len__`, `__contains__`, etc. This is the moment to add small public accessors (e.g., `iter_entries() -> Iterable[tuple[Candidate, EvalResult|None]]`, or `candidates`/`results` properties returning read-only views) instead of coupling `evolution.py` to frontier internals. Future refactors of `ParetoFrontier` storage will silently break this builder otherwise.

- **`HelixResult.budget: Any`.** `BudgetState` is a real dataclass in `state.py` and is the only thing ever stored here. Type it as `BudgetState` (importing from `helix.state`). The `to_dict()` `hasattr`/`getattr` fallback can then be replaced with `dataclasses.asdict(self.budget)` (or a dedicated `BudgetState.to_dict()`), which is more robust to future fields than the hardcoded 4-key whitelist that will silently drop newly-added budget fields.

- **`frontier_type` serialization.** `FrontierType` is a `Literal[str]`, so JSON is fine. Just confirming — no change needed.

- **Ordering inconsistency.** `non_dominated_ids = sorted(...)` is deterministic, but `frontier_ids = list(state.frontier)` preserves insertion order. Pick one (insertion-order is usually more useful for users; if you want both, expose `frontier_ids_sorted`). Mixing hides bugs in tests when iteration order changes.

- **Duplicate score storage.** `aggregate_scores`/`sum_scores`/`instance_scores`/`objective_scores` are stored both at the top level and inside each `CandidateSummary`. For large frontiers this doubles memory and creates two sources of truth that must agree. Consider either dropping the top-level dicts (let users do `{s.id: s.aggregate_score for s in result.candidate_summaries}`) or marking the per-summary copies as the canonical source. At minimum, document which is authoritative.

- **`parents` is duplicated similarly** (top-level dict + `CandidateSummary.parents`). Same observation.

- **Parent-resolution priority.** The fallback chain — `entry.parents` → `candidate.parent_ids` → `[entry.parent]` — is correct, but worth a one-line comment explaining why both lineage entry and candidate are consulted (legacy single-parent vs. merge).

- **`_build_helix_result` does I/O.** It calls `load_lineage(lineage_path)` which re-reads the lineage JSON we just finished writing. Acceptable at end-of-run, but if `_run_evolution_impl` already has the lineage in memory, prefer passing it in to avoid race conditions on long-running parallel callers and to make this function pure-ish.

- **Mutability of returned result.** `HelixResult` is a plain `@dataclass` with mutable list/dict fields. Users will mutate them. If that's a concern, use `field(default_factory=...)` and document immutability, or freeze with `frozen=True` + `tuple` collections.

---

### Tests

- **Coverage.** The new `test_returns_structured_helix_result` is thorough on the mutate path (good — it asserts every important field including `discovery_counts`, `seed`, `config_hash`, `frontier_type`, `to_dict()`). Gaps:
  - No test for the **merge** operation path through `_build_helix_result` (multi-parent → `parents` length > 1).
  - No test for **multiple non-dominated** candidates (currently only `g1-s1` is non-dominated). The sort/dedupe in `non_dominated_ids` and `is_non_dominated` flags are untested for >1 entry.
  - `HelixResult.to_dict()` itself isn't called/asserted on the full object — only `summaries[-1].to_dict()`. A round-trip through `json.dumps(result.to_dict())` would catch any non-JSON-serializable field (e.g., a future enum-typed `frontier_type`).

- **`make_config(frontier_type="instance")` blanket addition.** This change touches every existing evolution test. Why was it required now? If `EvolutionConfig`'s default already resolves to `"instance"` (state.py:111 suggests it does), this addition is noise; if the default changed in this PR, it's not in the diff and should be called out. Either revert these test edits or document the reason in the PR body.

- **`test_sandboxed_run_starts_evaluator_sidecar`** got the same `frontier_type="instance"` injection. Same comment.

---

### Performance

- For runs with thousands of candidates, building `HelixResult` materializes 4+ full dicts and a list of full `CandidateSummary` objects with copies of every per-instance score. That's O(N·M) memory at end-of-run. Consider lazy properties or `cached_property` accessors that build dicts on demand — useful for CLI runs that don't actually consume the structured result.

---

### Security
Nothing concerning. No new untrusted-input parsing, no new subprocess/network surface. `run_dir` is stored as `str(base_dir)` — fine; no path normalization needed at this layer.

---

### Nits
- Docstring of `_build_helix_result` is one line. A couple of lines on what's loaded from lineage vs. frontier vs. state would help future readers.
- `summaries.sort(key=lambda s: (s.generation, s.id))` is good. Worth noting in the dataclass docstring that the list is sorted by `(generation, id)` so users don't sort again.
- `objective_scores` ternary in the summary builder is repeated for the `objective_scores` top-level dict. A tiny `_clone_objective_scores(result)` helper would dedupe.

---

### Recommendation
**Approve with changes** once the following are addressed before merging out of draft:

1. (Must) Add public accessors to `ParetoFrontier` and stop touching `_candidates`/`_results` from `evolution.py`.
2. (Must) Type `HelixResult.budget` as `BudgetState` and replace the `to_dict()` whitelist with `asdict`.
3. (Should) Justify or revert the blanket `frontier_type="instance"` addition in `make_config` and the sandbox test.
4. (Should) Add a merge-path test and a multi-non-dominated test, plus a JSON-serialization round-trip on `HelixResult.to_dict()`.
5. (Nice) Note the API break beyond `best.id` in CHANGELOG/PR body and re-export `HelixResult`/`CandidateSummary` from `helix/__init__.py`.

Nice work overall — the structure mirrors upstream reference cleanly and the test exercises most of the surface.

KE7:
## Review feedback addressed in 24ec5e5

Pushed a follow-up commit on `fix/upstream-result-object` that addresses the items from the previous review.

### Changes
- **Dropped `HelixResult.id`** back-compat alias (per request — better to break than half-cover). Callers must use `result.best_candidate.id`. Updated the two in-tree assertions in `test_accepted_mutation_updates_frontier` and the align-iteration acceptance test.
- **Public `ParetoFrontier` accessors.** Added `candidates` and `results` `MappingProxyType` properties plus `get_result(cid)`. `_build_helix_result` no longer reaches into `_candidates` / `_results`.
- **`HelixResult.budget: BudgetState`** (was `Any`). `to_dict()` now uses `dataclasses.asdict(budget)` instead of a 4-key whitelist, so any future `BudgetState` field is preserved automatically.
- **Single source of truth for per-candidate data.** `parents`, `aggregate_scores`, `sum_scores`, `instance_scores`, `objective_scores` are now `@cached_property` views derived from `candidate_summaries`. The dataclass stores summaries once; the dict views are computed on first access. Same call-site ergonomics for users.
- **Documented ordering intent** inline: `frontier_ids` is insertion order from `state.frontier`; `non_dominated_ids` is sorted for stable diffing.
- **Documented `_build_helix_result` lineage read** at end-of-run (relies on synchronous `record_entry` flushes for authoritativeness) and the parent-resolution fallback chain (`LineageEntry.parents` → `Candidate.parent_ids` → legacy `LineageEntry.parent`).
- **Reverted blanket `frontier_type="instance"`** in `make_config` and the sandbox sidecar test. The arg is now optional; only the three tests that assert on `frontier_type` pass it explicitly.
- **Re-exported** `HelixResult`, `CandidateSummary`, `Candidate`, `EvalResult` from `helix.__init__` so callers can `from helix import HelixResult`.

### New tests (`tests/unit/test_evolution.py`)
- `test_helix_result_keeps_multiple_non_dominated_candidates` — two children with disjoint per-instance wins both flagged `is_non_dominated`; verifies `non_dominated_ids` is sorted.
- `test_helix_result_to_dict_round_trips_through_json` — `json.loads(json.dumps(result.to_dict()))` succeeds and budget fields survive intact (locks in `asdict` behavior; would catch any future enum/path/dataclass field that breaks JSON serializability).
- `test_helix_result_records_merge_parents` — exercises the multi-parent branch in `_build_helix_result`: a merged candidate's `parents` is `["g0-s0", "g1-s1"]` and its `CandidateSummary.operation == "merge"`.

### Verification
- `pytest`: **687 passed** (3 new tests added on top of the prior 684).
- `ruff check`: clean on touched files.
- `mypy src/helix/{population,evolution,__init__}.py`: clean.

### Items intentionally not addressed
- The pre-existing `frontier._results` / `frontier._candidates` accesses elsewhere in `evolution.py` (lines 1294, 1402, 1412, 1447–1489, 1742, 1832, 2475, 2516) and the `render_frontier_table(frontier, frontier._results)` callsite are out of scope for this PR — they pre-date `fix/upstream-result-object`. Recommend a follow-up to switch them over to the new public accessors.
- Pre-existing `F841` ruff issue in `examples/web_researcher/evaluate.py:248` is on the branch but unrelated to this PR.

Lifting out of draft is recommended once you re-review.

KE7:
## Merged `main` into `fix/upstream-result-object` (b95eb9e)

PR is now `MERGEABLE` / `CLEAN` again.

### Conflicts resolved
- **`src/helix/__init__.py`** — combined the new `from helix.reserved metadata import log` (main, PR #21) with the new dataclass re-exports (this PR). `__all__` now lists `Candidate`, `CandidateSummary`, `EvalResult`, `HelixResult`, `__version__`, `log`.
- **`src/helix/population.py` (imports)** — main moved `Iterator` to `collections.abc`; this PR added `cached_property`, `MappingProxyType`, `asdict`, and `Mapping`. Combined: `Iterator` and `Mapping` both come from `collections.abc`; the rest stay where they were.
- **`src/helix/population.py` (`ParetoFrontier` accessors)** — both branches independently added a public read-only API. Main added `get_result(cid)`, `iter_results()`, and `has_results()`; this PR added `candidates` and `results` `MappingProxyType` properties plus `get_result(cid)`. The `get_result` definitions were identical (modulo docstring nit). Final API exposes all five: `candidates`, `results`, `get_result`, `iter_results`, `has_results`.
- **`tests/unit/test_evolution.py` (imports)** — combined main's `_real_generate_seed` import with this PR's `HelixResult` import.

### Test fix-up post-merge
- `test_returns_structured_helix_result` previously asserted `discovery_counts == {"g0-s0": 1, "g1-s1": 3}` and `result.budget.evaluations == 3`. Main's PR #16 (centralized budget accounting) bumped the `g1-s1` count to 4. The test now locks in the *invariants* — discovery monotonicity and `discovery_counts[best] == budget.evaluations` — which survive future budget-policy changes:
  ```python
  assert set(result.discovery_counts) == {"g0-s0", "g1-s1"}
  assert result.discovery_counts["g0-s0"] < result.discovery_counts["g1-s1"]
  assert result.discovery_counts["g1-s1"] == result.budget.evaluations
  ```

### Verification
- Full pytest: **748 passed** (was 687 pre-merge — main brought in PR #16 / PR #21 / PR #22 budget+reserved metadata+cache tests).
- `ruff check` on touched files: clean.
- `mypy` on `src/helix/{__init__,population,evolution}.py`: clean.

Ready to lift out of draft and merge once you re-review.

KE7:
## upstream reference-parity immutability shape on `HelixResult` (e1e7233)

Following the discussion on whether to use tuples vs `frozen=True`, I went and read upstream reference actual `result` (`upstream reference-ai/upstream reference` → `src/upstream reference/core/result.py`) — they use `@dataclass(frozen=True)` with plain `list`/`dict`/`set` fields and shallow defensive copies in `from_state`.  Adopted the same shape here for parity.

### Changes
1. **`HelixResult` is now `@dataclass(frozen=True)`.** Top-level attribute reassignment raises `FrozenInstanceError` — same guarantee `result` provides. Inner collections stay plain `list`/`dict` (same as upstream reference — they explicitly chose shallow-immutability over `tuple`/`MappingProxyType` for ergonomics).
2. **`@cached_property` → `@property` for the five score views** (`parents`, `aggregate_scores`, `sum_scores`, `instance_scores`, `objective_scores`). Required by `frozen=True` (cached_property writes to `__dict__`) and — more importantly — eliminates the cache-staleness window where a view could disagree with `candidate_summaries` after any mutation. Recompute cost is a microsecond dict-comp per access; `result.best_idx` accepts the same trade.
3. **`_build_helix_result` shallow-copies `budget`** via `dataclasses.replace(state.budget)`. The previous `budget=state.budget` aliased the live `BudgetState` that the budget-accounting code mutates throughout the run; the snapshot now stands alone. Brings `budget` in line with the existing `list(...)` / `dict(...)` defensive copies on every other collection field — and matches `result.from_state` discipline.

### Three new regression tests (`tests/unit/test_evolution.py`)
- `test_helix_result_is_frozen` — `result.seed = 99` raises `FrozenInstanceError`.
- `test_helix_result_budget_does_not_alias_state` — verifies the budget is decoupled and `result.budget = ...` reassignment is blocked.
- `test_helix_result_score_views_recompute_from_summaries` — clears `candidate_summaries`, asserts the score views immediately reflect `{}` (no stale cache), restores, asserts they recompute back.

### Verification
- Full pytest: **751 passed** (was 748; +3 new tests).
- `ruff check` on touched files: clean.
- `mypy src/helix/{population,evolution}.py`: clean.

### What this *doesn't* do (intentional)
- Inner `list`/`dict` fields are still mutable in place — `result.candidate_summaries.append(...)` will not raise. This matches `result` exactly. Going further (tuple, `MappingProxyType`, deep freeze) would diverge from upstream reference and isn't worth the API tax.
- `best_result` (the per-`EvalResult`) is still aliased into the frontier's storage. `EvalResult` isn't mutated post-run by anything in HELIX, and `result` makes the same trade for its analogous fields. Can revisit if a real bug shows up.

`HelixResult.id` stays as-is (your call; it's the run's logical identifier and the most-accessed attribute by a wide margin — `frozen=True` doesn't affect properties).

PR #20, opened 2026-05-07
Enforce upstream reference frontier and acceptance semantics

KE7:
## Summary
- Makes non-instance frontier modes fail loudly when objective scores are missing or malformed.
- Adds MissingObjectiveScoresError and stricter helix_result objective-score validation.
- Keeps all candidates distinct from the active frontier snapshot.
- Preserves strict upstream reference multi-axis behavior for objective, hybrid, and cartesian frontiers.

## Validation
- Clean branch E2E/parser/frontier suite: 75 passed.
- Clean branch full unit suite: 688 passed.
- Worker verification also reported tests pass except full repo ruff blocked by pre-existing examples/web_researcher/evaluate.py F841.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   validation errors, pointing users at likely typos or misplaced keys.
 
 ### Changed
+- **BREAKING**: `score_parser = "helix_result"` now rejects malformed
+  `side_info["scores"]` payloads instead of silently dropping fields.
+  Non-dict `"scores"` values, non-string objective names, and non-finite
+  / non-numeric objective values now raise `EvaluatorError` at parse
+  time.  Previously these were dropped (with a logged warning at most),
+  which let evaluators stuff arbitrary diagnostics under `"scores"` —
+  including pairwise / Bradley-Terry payloads — and then misread the
+  result as scalar objective semantics.  Pairwise / BT payloads are
+  not implemented in HELIX yet; emit them under a different
+  `side_info` key.  Migration: ensure `side_info["scores"]` is a
+  `dict[str, float|int|bool]` with finite numeric values, or omit the
+  key entirely.
+- **BREAKING**: Non-`"instance"` `evolution.frontier_type` values now
+  fail loudly when an `EvalResult` lacks per-example `objective_scores`,
+  has length-mismatched `objective_scores`, or has all-empty objective
+  slots.  `ParetoFrontier.add()` raises the new
+  `MissingObjectiveScoresError` instead of silently degenerating to
+  scalar / instance-axis behaviour.  `select_parent()` likewise raises
+  rather than falling back to all-candidate sampling on objective-bearing
+  modes.  The `"instance"` path keeps its existing fallback semantics.
 - **BREAKING**: `score_parser = "helix_result"` now takes a **per-example**
   `HELIX_RESULT=[[score_0, side_info_0], [score_1, side_info_1], ...]`
   payload — one `[score, side_info]` pair per id in `helix_batch.json`.
--- a/src/helix/config.py
+++ b/src/helix/config.py
@@ -95,8 +95,8 @@ class EvaluatorConfig(BaseModel):
     ``"json_score"``) aggregate to a single score and do NOT produce
     id-keyed per-instance scores; combining them with ``instance_ids``
     triggers the zero-fill warning in :mod:`helix.executor`.  They also
-    leave ``objective_scores`` empty, so the multi-axis frontier
-    degenerates to the instance path when they're selected.
+    leave ``objective_scores`` empty, so ``frontier_type`` values other
+    than ``"instance"`` fail loudly when selected with these parsers.
     """
 
     model_config = ConfigDict(extra="forbid")
@@ -410,8 +410,8 @@ class EvolutionConfig(BaseModel):
             "only the Pareto retention / parent-selection decision is "
             "multi-axis.  Non-``instance`` paths require ``helix_result`` "
             'to emit per-example ``side_info["scores"]`` dicts — without '
-            "them the objective / cartesian frontiers stay empty and "
-            "behaviour degenerates to the instance path."
+            "them the objective / cartesian frontiers raise instead of "
+            "falling back to scalar semantics."
         ),
     )
 
--- a/src/helix/evolution.py
+++ b/src/helix/evolution.py
@@ -1058,6 +1058,11 @@ def _run_evolution_impl(
         if minibatch_cache is not None:
             save_eval_cache(minibatch_cache._cache, project_root)
 
+    def _sync_frontier_state() -> None:
+        assert state is not None
+        state.frontier = list(frontier._candidates.keys())
+        state.active_frontier = frontier.active_frontier_snapshot()
+
     if state is None:
         state = EvolutionState(
             generation=0,
@@ -1102,6 +1107,7 @@ def _run_evolution_impl(
                 )
                 candidates[cid] = cand
                 frontier.add(cand, result)
+        _sync_frontier_state()
 
     # ------------------------------------------------------------------
     # Seed evaluation
@@ -1196,7 +1202,7 @@ def _run_evolution_impl(
 
         _save_evaluation(base_dir, seed_result)
         frontier.add(seed, seed_result)
-        state.frontier = list(frontier._candidates.keys())
+        _sync_frontier_state()
         state.instance_scores[seed.id] = seed_result.instance_scores
         # GEPA parity (audit-rng-state-persist C/§3): record per-program
         # discovery budget at the moment the program enters the frontier.
@@ -1644,7 +1650,7 @@ def _run_evolution_impl(
                                 merges_due -= 1
                                 state.total_merge_invocations += 1
                                 frontier.add(merged, full_val_result)
-                                state.frontier = list(frontier._candidates.keys())
+                                _sync_frontier_state()
                                 state.instance_scores[merged.id] = (
                                     full_val_result.instance_scores
                                 )
@@ -2377,7 +2383,7 @@ def _run_evolution_impl(
                 set_phase(HelixPhase.PARETO_UPDATE)
                 _save_evaluation(base_dir, val_result)
                 frontier.add(child, val_result)
-                state.frontier = list(frontier._candidates.keys())
+                _sync_frontier_state()
                 state.instance_scores[child.id] = val_result.instance_scores
                 # GEPA parity (audit-rng-state-persist C/§3): record per-program
                 # discovery budget at the moment the child enters the frontier.
--- a/src/helix/evolution.py
+++ b/src/helix/evolution.py
@@ -1060,7 +1060,7 @@ def _run_evolution_impl(
 
     def _sync_frontier_state() -> None:
         assert state is not None
-        state.frontier = list(frontier._candidates.keys())
+        state.frontier = frontier.candidate_ids()
         state.active_frontier = frontier.active_frontier_snapshot()
 
     if state is None:
--- a/src/helix/parsers/helix_result.py
+++ b/src/helix/parsers/helix_result.py
@@ -147,6 +147,25 @@ def _harvest_objective_scores(side_info: dict[str, Any]) -> dict[str, float]:
     not implemented in HELIX yet, so non-scalar objective values fail
     loudly instead of being dropped and misread as scalar objective
     semantics.
+
+    Stricter than upstream GEPA (a very minor improvement)
+    ------------------------------------------------------
+    GEPA's adapter does ``objective_score.update(side_info["scores"])``
+    blindly, so non-finite / non-numeric / non-string-keyed entries
+    sail through the parser and only blow up later inside
+    ``_aggregate_objective_scores`` (``state.py:432``) where the error
+    message is far less actionable.  HELIX validates at parse time and
+    raises :class:`EvaluatorError` with the offending key/value type.
+    Practical consequence: an evaluator built against upstream GEPA
+    that legitimately emits finite numeric scalars under ``"scores"``
+    is unaffected; only payloads that would break GEPA later are
+    rejected here earlier.
+
+    Per-predictor namespacing (``param_name + "_specific_info"``,
+    ``optimize_anything_adapter.py:266-270``) is intentionally not
+    replicated: HELIX evolves whole git worktrees rather than
+    multi-component named-predictor programs.  See the tracking issue
+    on ``KE7/helix`` for the architectural gap.
     """
     raw = side_info.get("scores")
     if raw is None:
--- a/src/helix/parsers/helix_result.py
+++ b/src/helix/parsers/helix_result.py
@@ -143,24 +143,40 @@ def _harvest_objective_scores(side_info: dict[str, Any]) -> dict[str, float]:
     (``optimize_anything_adapter.py:260-272``): the reserved
     ``"scores"`` key carries a dict of ``{objective_name: float}`` that
     becomes the per-example ``objective_scores`` slot feeding the
-    multi-axis Pareto frontier.  Non-dict values and non-numeric /
-    non-finite entries are silently dropped so evaluators can stuff
-    arbitrary diagnostics under ``"scores"`` without the parser
-    rejecting well-formed payloads elsewhere (the primary per-example
-    score is what the minibatch gate depends on; the objective axis
-    is best-effort).
+    multi-axis Pareto frontier.  Pairwise / Bradley-Terry payloads are
+    not implemented in HELIX yet, so non-scalar objective values fail
+    loudly instead of being dropped and misread as scalar objective
+    semantics.
     """
     raw = side_info.get("scores")
-    if not isinstance(raw, dict):
+    if raw is None:
         return {}
+    if not isinstance(raw, dict):
+        raise EvaluatorError(
+            'helix_result side_info["scores"] must be a dict of '
+            "{objective_name: numeric_score}; pairwise/Bradley-Terry "
+            "objective payloads are not supported.",
+            operation="parse_helix_result",
+        )
     out: dict[str, float] = {}
     for k, v in raw.items():
         if not isinstance(k, str):
-            continue
+            raise EvaluatorError(
+                'helix_result side_info["scores"] objective names must be strings; '
+                f"got {type(k).__name__}.",
+                operation="parse_helix_result",
+            )
         if isinstance(v, bool):
             out[k] = 1.0 if v else 0.0
         elif isinstance(v, (int, float)) and math.isfinite(float(v)):
             out[k] = float(v)
+        else:
+            raise EvaluatorError(
+                'helix_result side_info["scores"] values must be finite numeric '
+                f"scalars; objective {k!r} had {type(v).__name__}. "
+                "Pairwise/Bradley-Terry objective payloads are not supported.",
+                operation="parse_helix_result",
+            )
     return out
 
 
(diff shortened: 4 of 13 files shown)


KE7:
@claude review

KE7:
## Code Review — PR #20: Enforce upstream reference frontier and acceptance semantics

### Overview
Tightens the multi-axis Pareto frontier path so non-`instance` `frontier_type` values fail loudly (rather than silently degenerating to scalar/instance semantics) when objective scores are missing or malformed. Adds:

- `MissingObjectiveScoresError` raised in `ParetoFrontier.add()` / `select_parent()` / `_update_cartesian()`.
- Strict validation in the `helix_result` parser's `_harvest_objective_scores`: non-dict `scores`, non-string keys, and non-finite/non-numeric values now raise `EvaluatorError` instead of being dropped.
- A new `state.active_frontier` JSON-native snapshot (kept distinct from the append-only `state.frontier` candidate-id list), persisted via `_sync_frontier_state()` at every accept site (init, seed, mutation, merge). `SCHEMA_VERSION` bumped 1 → 2.
- Test refactors: parametrized fail-loud tests on the multi-axis frontier; per-call `frontier_type="instance"` pinning in legacy evolution tests.

### Strengths
- **Semantics match upstream reference parity comments** in the existing code; the change closes the silent-fallback footgun called out in the docstrings.
- **Loader migration is forward/backward safe**: `version > SCHEMA_VERSION` only rejects newer states, and missing fields default via `data.get(...)`. v1 → v2 will load cleanly.
- `MissingObjectiveScoresError(ValueError)` keeps callers that catch `ValueError` working.
- `active_frontier_snapshot()` is a deterministic (sorted) deep copy — good for diffable JSON state.
- The new validation triplet (`is None` / length mismatch / all-empty slots) covers the realistic failure modes.
- Test coverage for the new error paths is solid (parametrized across `objective`/`hybrid`/`cartesian`, plus pairwise/BT-payload rejection).

### Issues / Suggestions

**1. Indentation bug in `tests/unit/test_evolution.py` (line ~1734).** The edit slipped one indent level:

```python
        sandbox=SandboxConfig(
            enabled=True,
            ...
        ),
            evolution=EvolutionConfig(max_generations=0, frontier_type="instance"),
        )
```
`evolution=...` is indented 12 spaces while sibling kwargs (`evaluator=`, `sandbox=`) are at 8, and the closing `)` is at 8. It parses, but it's inconsistent with the rest of the file and will trip ruff/format checks. Please re-indent to match.

**2. `update_scores()` skips the new validation.** `ParetoFrontier.add()` calls `_validate_objective_scores`, but `update_scores()` calls `_rebuild_axes()` directly. A caller that overwrites an existing candidate's `EvalResult` with a stripped/objective-less result would bypass the new invariant. Either:

- call `_validate_objective_scores(result.candidate_id, result)` at the top of `update_scores()`, or
- document explicitly that `update_scores` assumes invariants hold.

**3. Length-check duplication in `_update_cartesian`.** With validation now in `add()`, the length-mismatch raise inside `_update_cartesian` is unreachable on the `add()` path and only fires from `_rebuild_axes()` (i.e., post-`update_scores`). That's still useful as defense-in-depth, but worth a one-line comment so it doesn't read as dead code.

**4. Error class name vs. usage.** `MissingObjectiveScoresError` is also raised for *length mismatch* and *empty active frontier in `select_parent`*. Both are technically "objective coverage missing", but a reader scanning for `Missing...` may not expect the length-mismatch case. Consider either widening the docstring on the class or splitting into `InvalidObjectiveScoresError` for shape errors. (Cosmetic.)

**5. `_sync_frontier_state()` reaches into `frontier._candidates`.** Same pattern as existing code, but if you're already adding `active_frontier_snapshot()`, consider a sibling `candidate_ids()` property to avoid touching the private attribute from `evolution.py`. Optional cleanup.

**6. Redundant `frontier_type="instance"` in tests?** `EvolutionConfig.frontier_type` already defaults to `"instance"` — the explicit pinning works as nice belt-and-suspenders against future default changes, but a one-line comment in the most-edited file (e.g., `test_evolution_minibatch.py`) explaining *why* (`# pin: tests build EvalResults without objective_scores`) would help future readers.

**7. `_update_objective` early-return.** Still bails on `objective_scores is None or not objective_scores`. The validator now rejects both cases for non-`instance` modes before this is reached, but the helper is still called for `instance` mode (where objectives are intentionally optional). Behavior is correct; just call out in the comment that the early-return is the `instance`-mode path now.

### Correctness / Risk
- No silent behavior changes for the default `instance` path — existing runs unaffected.
- Schema migration v1 → v2 verified: the `active_frontier` field defaults to `{}` for legacy state files.
- The parser change (`raw is None` → `{}`, anything else dict-or-raise) is correct: upstream reference adapter only ever emits a dict here, and absent `"scores"` is the common "no objectives configured" case.
- One behavior change worth flagging in release notes: evaluators that previously emitted `side_info["scores"]` with stray non-numeric junk (e.g. NaN diagnostics) will now hard-fail at parse time. PR description should call this out as a breaking change for evaluator authors.

### Tests
- Frontier coverage: good (added parametrized + cartesian + hybrid empty-slot cases).
- Parser coverage: good (non-dict, non-finite, pairwise BT payload).
- State persistence: roundtrip + json-keys updated for `active_frontier`. ✓
- Missing: an evolution-level integration test that runs `frontier_type="objective"` end-to-end and confirms `state.active_frontier` is populated post-`_save_state`. The new sync points are touched in 4 places — a smoke test guarding the call sites would prevent regressions if a future accept site is added without the sync call.

### Minor / Nits
- `population.py` removed the `logging` import along with the debug log — clean.
- The PR title says "acceptance semantics" but no acceptance-criterion code is changed in the diff (only frontier-side rejection). If acceptance was tightened elsewhere on this branch, the diff shown here doesn't include it; otherwise consider retitling to just "Enforce upstream reference frontier semantics".

### Verdict
Solid hardening change with the right "fail loud, don't silently degrade" instinct. Address (1) before merge; (2) is the only other one I'd consider close-to-blocking. Everything else is polish.

KE7:
## Review follow-ups addressed in 1033233

All 7 review issues addressed in commit `1033233`:

| # | Issue | Resolution |
|---|---|---|
| 1 | Indentation bug in `test_evolution.py` ~L1734 | Re-aligned `evolution=` kwarg and closing `)` to match siblings |
| 2 | `update_scores()` bypassed new validation | Added `_validate_objective_scores` call at the top of `update_scores` |
| 3 | Length-check duplication in `_update_cartesian` | Inline comment marks it as defense-in-depth for the `_rebuild_axes` path |
| 4 | `MissingObjectiveScoresError` name vs. usage | Widened docstring to enumerate all "objective axis unusable" shapes |
| 5 | `_sync_frontier_state` reached into `frontier._candidates` | Added public `ParetoFrontier.candidate_ids()` and routed `_sync_frontier_state` through it |
| 6 | Breaking change for evaluator authors not flagged | Added two **BREAKING** entries to `CHANGELOG.md` (parser rejection of malformed `side_info["scores"]`; `MissingObjectiveScoresError` from non-`instance` frontier modes) |
| 7 | Missing objective-mode integration test | Added `TestActiveFrontierSyncOnObjectiveMode.test_seed_acceptance_populates_active_frontier_for_objective_mode` — drives `run_evolution` end-to-end with `frontier_type="objective"` and asserts `state.active_frontier == {"quality": ["g0-s0"]}` after seed acceptance |

### Verification
- Full unit suite: **689 passed** (was 688; +1 for the new integration test).
- `ruff check` clean on all touched files.

KE7:
## upstream reference parity re-verification

Re-reviewed PR #20 (now at `5b41a15` after main merge) against upstream `upstream reference-ai/upstream reference` HEAD.  Anchoring against:

- `src/upstream reference/core/state.py` (validation + `_update_*` helpers)
- `src/upstream reference/optimize_anything.py` (defaults)
- `src/upstream reference/adapters/optimize_anything_adapter/optimize_anything_adapter.py` (side_info → objective_scores extraction)
- `src/upstream reference/upstream reference_utils.py` (`select_program_candidate_from_pareto_front`)

### ✅ Confirmed parity

| HELIX site | upstream reference analogue | Notes |
|---|---|---|
| `EvolutionConfig.frontier_type = "hybrid"` (default) | `EngineConfig.frontier_type: FrontierType = "hybrid"` (`optimize_anything.py:476`) | ✓ matches upstream reference default |
| `EvolutionConfig.cache_evaluation = False` (default, post-merge) | `EngineConfig.cache_evaluation: bool = False` (`optimize_anything.py:498`) | ✓ matches upstream reference default |
| `ParetoFrontier._update_objective` (mean-across-val_id, then `> / ==` per-objective set-add) | `_update_objective_pareto_front` (`state.py:474-484`) | ✓ semantics line up; HELIX aggregates inside the helper instead of pre-aggregating, but result is identical |
| `ParetoFrontier._update_cartesian` (per-`(val_id, obj_name)` max, ties set-add) | `_update_pareto_front_for_cartesian` (`state.py:512-525`) | ✓ HELIX flattens the tuple key into `f"{val_id}::{obj_name}"` so the existing `_remove_dominated_programs(dict[str, set[str]])` helper works unchanged. Functionally equivalent. |
| `ParetoFrontier._validate_objective_scores` raises at every `add()` | upstream reference validates at construction (`state.py:213-219`) AND every `update_state_with_new_program` (`state.py:564-569`) | ✓ same touchpoint coverage. HELIX raises a named `MissingObjectiveScoresError(ValueError)` subclass; upstream reference raises plain `ValueError`. HELIX's named subclass is strictly more catchable — backwards-compatible. |
| `select_parent()` raises on empty active frontier (non-instance modes) | `select_program_candidate_from_pareto_front` does `assert len(sampling_list) > 0` (`upstream reference_utils.py:114`) | ✓ both fail loud; HELIX raises a typed exception, upstream reference throws `AssertionError`. HELIX's typed error is preferable. |
| `EvalResult.objective_scores: list[dict[str, float]] \| None`, positional to instance id order | `EvaluationBatch.objective_scores: Sequence[ObjectiveScores] \| None` (`adapter.py:26`) | ✓ same shape, positional contract preserved |
| Bool coercion in `_harvest_objective_scores` (`True → 1.0`, `False → 0.0`) | upstream reference `objective_score.update(side_info["scores"])` (`optimize_anything_adapter.py:264`) followed by `totals[obj] += score` (`state.py:432`) — relies on Python implicit `True + 0.0 == 1.0` | ✓ HELIX is *explicit* about the coercion; numerical result is identical to upstream reference |
| `state.active_frontier` snapshot persisted alongside append-only `state.frontier` | upstream reference pickles its full state including `program_at_pareto_front_*` (`state.py:306-340`) | ✓ HELIX's split (`frontier` = append-only ids, `active_frontier` = retained fronts) is the JSON-friendly equivalent of upstream reference pickle round-trip |

### ⚠️ Intentional / acceptable divergences

1. **Per-predictor objective namespacing not replicated.**
   upstream reference `_process_side_info` (`optimize_anything_adapter.py:266-270`) also harvests per-predictor scores from `side_info[param_name + "_specific_info"]["scores"]` and prefixes them as `f"{param_name}::{k}"`.  HELIX's `_harvest_objective_scores` only reads top-level `side_info["scores"]`.  This is correct: HELIX has no multi-predictor / `named_predictors` concept (verified — `grep -r 'named_predictor\|specific_info' src/helix/` returns zero hits).  The simpler harvest is the right level of fidelity.
   **Suggestion:** Add a one-line note in the parser docstring stating that the upstream reference per-predictor namespacing is intentionally omitted because HELIX evolves whole worktrees, not named-predictor programs.

2. **HELIX's parser is *stricter* than upstream reference adapter.**
   upstream reference does `objective_score.update(side_info["scores"])` blindly — non-finite, NaN, str, dict-of-dict (BT/pairwise) all sail through and reach `_aggregate_objective_scores` where they break later (e.g. `NaN + 0.0 = NaN`).  HELIX raises `EvaluatorError` at parse time.  Strictly an improvement, but it does mean an evaluator that worked against upstream upstream reference can hit `EvaluatorError` here.  The CHANGELOG entry already documents this.

3. **HELIX's "all-empty slots" check has no upstream reference analogue.**
   upstream reference seed-time check is `if not valset_evaluation.objective_scores_by_val_id` — i.e. only rejects when the *outer* dict is empty/None.  upstream reference tolerates the case where every per-val_id inner dict is `{}` (its `_update_objective_pareto_front` has `if not objective_scores: return` to handle the resulting empty aggregation).  HELIX's `_validate_objective_scores` adds:
   ```python
   if not any(slot for slot in result.objective_scores):
       raise MissingObjectiveScoresError(...)
   ```
   which **rejects all-empty inner slots that upstream reference would silently no-op**.  Likely intentional ("if no signal exists, fail loud") and consistent with the PR's posture, but it does mean an evaluator that emits sparse objective scores — ALL empty on a particular batch — will hit a hard failure instead of a no-op.

   **Suggestion:** Either (a) document this as a deliberate stricter-than-upstream reference invariant in the `_validate_objective_scores` docstring with a one-liner explaining why HELIX prefers the hard failure, or (b) downgrade to upstream reference posture and rely on the downstream "empty active frontier" raise in `select_parent`.  I'd lean (a) — the early failure is more diagnosable than a confusing empty-frontier raise three call sites later.

4. **HELIX's `len(objective_scores) == len(instance_scores)` check.**
   upstream reference stores objective scores as a `dict[val_id, ObjectiveScores]` keyed parallel to `scores_by_val_id`, so the cardinality match is a structural impossibility (same dict keyspace).  HELIX uses *positional* `list` against `instance_scores.keys()` order — meaning the length check is the necessary structural guard.  ✓ Justified.

### Quality of the new test
`TestActiveFrontierSyncOnObjectiveMode.test_seed_acceptance_populates_active_frontier_for_objective_mode` exercises only the *seed-acceptance* sync site (line 1205 in `evolution.py`).  The other three sync sites (init/load at 1110, merge at 1653, mutation at 2386) are not covered by the new test.  Existing per-touchsite tests aren't affected because they pin `frontier_type="instance"`.

**Suggestion:** Either parametrize the new test across the four sync points (add a mutation-accept variant), or rely on a coarser invariant test that drives a 1-generation run with one mutation under `frontier_type="objective"`.  Not blocking — the seed path is the most-traveled and the others share `_sync_frontier_state`.

### Verdict
upstream reference parity holds on the core semantics (frontier updates, validation gating, defaults, snapshot shape).  The two stricter-than-upstream reference behaviors (parser validation, all-empty-slots) are arguably improvements but should be acknowledged in code comments so a future maintainer doesn't "fix" them back to upstream reference looser semantics.

KE7:
## Conflicts resolved + upstream reference parity re-verified

Re-merged `main` twice (PR #21 reserved metadata-log-channel landed first, then PR #19 helix-result immutability landed during my push) — branch is now mergeable at `b5722ba`.

**Resolutions:**
- `CHANGELOG.md`: kept both `Changed` blocks (cache_evaluation default + this PR's two BREAKING entries).
- `src/helix/population.py`: kept the new `TYPE_CHECKING: BudgetState` import from PR #19; dropped the `logger`/`import logging` since this PR replaced the only `logger.debug` site (cartesian length-mismatch) with a `MissingObjectiveScoresError` raise.
- `tests/unit/test_evolution.py::make_config`: reconciled main's `frontier_type` parameterisation (default `None` → `"hybrid"` via `EvolutionConfig`) with this PR's inline `"instance"` pin.  The helper's default is now `frontier_type="instance"` because the bulk of the suite drives `run_evolution` against `make_eval_result` outputs that carry no `objective_scores` — under this PR's stricter `_validate_objective_scores` they would raise on `"hybrid"`.  Multi-axis tests pass an explicit value.
- Pinned `frontier_type="instance"` on the two new `TestLlmUsageBudgetIntegration` seedless-budget tests (same reason).

**Verification:**
- 757 unit tests pass (was 750 before second merge; up from PR baseline 689).
- `ruff check` clean across `src/` and `tests/`.

A separate parity audit against `upstream reference-ai/upstream reference` HEAD (`optimize_anything.py`, `core/state.py`, `upstream reference_utils.py`, `optimize_anything_adapter.py`) is in https://github.com/KE7/helix/pull/20#issuecomment-4414708941 — TL;DR: core semantics line up; HELIX is intentionally stricter than upstream reference on parser validation and "all-empty-objective-slots", which is worth one short comment in the validator docstring.

KE7:
## Round-3 follow-ups (commit `d796f59`)

Addressed the four items raised on the parity-review thread:

### 1. Test defaults now match upstream reference
Production already defaults `frontier_type="hybrid"` and `cache_evaluation=False`.  The test helpers were pinning `"instance"`, masking any drift.  Now:

- `make_eval_result` (in `test_evolution.py` and `test_evolution_seedless.py`) synthesises a per-example `objective_scores` slot list (`[{"quality": <score>}, ...]`) by default so results pass `_validate_objective_scores` under `"hybrid"`.  Tests that intentionally hit the missing-objectives path pass `objective_scores=None` explicitly via a `_DEFAULT_OBJECTIVES` sentinel.
- `make_config` flips default `"instance"` → `"hybrid"`.
- All 8 inline `frontier_type="instance"` pins removed from `make_config(...)` / inline `EvolutionConfig(...)` call sites.
- The two `test_returns_structured_helix_result` / `test_helix_result_to_dict_round_trips_through_json` assertions on `result.frontier_type` flip from `"instance"` → `"hybrid"` to match.
- `test_evolution_minibatch.py` is intentionally untouched: its `_make_minibatch_config` is a specialised helper used by tests that hand-build varied `EvalResult` shapes to exercise minibatch internals.

### 2. Stricter-than-upstream reference divergences kept, now advertised as minor improvements
Both behaviours stay (they catch real bugs and align with the PR's fail-loud thesis), with explicit docstring callouts so future maintainers don't "fix" them back to upstream reference looser semantics:

- `helix_result._harvest_objective_scores` docstring: explicit comparison to upstream reference `objective_score.update(side_info["scores"])` (`optimize_anything_adapter.py:264`) — non-finite / non-numeric / non-string-keyed entries that would only blow up inside upstream reference `_aggregate_objective_scores` are caught here at parse time with the offending key/value type.  Frame: "very minor improvement that should never trigger for a well-formed evaluator".
- `ParetoFrontier._validate_objective_scores` docstring: lists the two narrow places where HELIX is *very slightly* stricter than `state.py:213-219` / `:564-569` — the length-parity check (necessary structural invariant for HELIX's positional-list shape; upstream reference stores objective scores as a parallel-keyed dict where this match is structural) and the all-empty-slot rejection (a candidate with literally zero objective signal cannot contribute to any objective frontier; failing here surfaces the empty-axis condition with a per-candidate error rather than the downstream "empty active frontier" raise three call sites later).

### 3. Multi-predictor parity gap tracked
Opened https://github.com/KE7/helix/issues/24 — covers the upstream reference **named-predictor / multi-component candidate** model (the missing per-predictor objective namespacing was the only remaining upstream reference parity gap surfaced in this review).  Issue includes a feasibility sketch, suggested first step (parser-side change behind an off-by-default config key), and references back to this PR thread.

### 4. `TestActiveFrontierSyncOnObjectiveMode` parametrised across 3 of 4 sync sites
Added two regression tests:

- `test_mutation_acceptance_updates_active_frontier_for_objective_mode` — runs one generation with a child whose mean objective strictly displaces the seed; asserts `active_frontier == {"quality": ["g1-s1"]}` at the post-mutation save call.
- `test_init_sync_rebuilds_active_frontier_on_resume` — mocks `load_state` to return a state with a stale `active_frontier`; uses `mocker.spy(ParetoFrontier, "active_frontier_snapshot")` to observe the rebuild directly (save-state plumbing isn't reachable when `max_generations=0`).

The merge sync site shares the same `_sync_frontier_state` closure and is exercised by the existing instance-mode merge tests; a parametrised version would require driving a full merge accept path under `frontier_type="objective"`, which is non-trivial and lower marginal value given the shared helper.

### Verification
- Full unit suite: **759 passed** (was 757).
- `ruff check` clean across `src/` and `tests/`.

PR #21, opened 2026-05-10
Add evaluator reserved metadata log channel

KE7:
## Summary
- Add a general HELIX reserved metadata log channel for evaluator-visible messages.
- Thread reserved metadata logging through executor and mutator flows.
- Add unit coverage for executor and mutator reserved metadata log behavior.

## Scope check
- Branch is based on `origin/main` and contains one cherry-picked commit from `dea0314` (`a4c81bb`).
- Diff is limited to general HELIX reserved metadata/log channel files.
- Checked the diff for CaP-X/cap-x/robosuite/R1Pro/LIBERO/domain-specific prompt-filter terms; no matches.

## Tests
- `.venv/bin/ruff check .` (fails on existing unrelated `F841` in `examples/web_researcher/evaluate.py:248`)
- `.venv/bin/ruff check README.md src/helix/__init__.py src/helix/reserved metadata.py src/helix/executor.py src/helix/mutator.py tests/unit/test_executor.py tests/unit/test_executor_security.py tests/unit/test_mutator.py`
- `.venv/bin/pytest tests/unit` (690 passed)

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,7 +330,8 @@ command = "uv run python evaluate.py"
 #   HELIX_RESULT=[s_0, [s_1, si_1], s_2, ...]           # mixed
 # Positional to `helix_batch.json`. HELIX zips it into id-keyed
 # `instance_scores` and stores the side_info list for the reflection
-# prompt. Use for minibatch runs (`dataset.train_size` set).
+# prompt. `HELIX_RESULT` is a machine protocol; use `from helix import log`
+# for human-readable evaluator notes.
 score_parser = "json_score"
 include_stdout = true
 include_stderr = true
@@ -564,6 +565,19 @@ print(json.dumps({
     "instance_scores": {"puzzle_001": 1.0, "puzzle_002": 0.0}
 }))
 ```
+
+For GEPA-style reflective feedback, prefer structured per-example
+`side_info` when using `score_parser = "helix_result"`. For additional
+human-readable notes that are not tied to one example, evaluators may call
+`from helix import log` and then `log("what happened", key=value)`. HELIX
+captures these notes through `HELIX_ASI_LOG`, not stdout, and renders them as
+Evaluator Notes in mutation prompts.
+
+Raw `HELIX_RESULT=...` JSON is only consumed by HELIX as a machine protocol and
+is not shown to mutators. Evaluator stdout/stderr remain stored for debugging,
+but successful mutation prompts omit them when structured diagnostics or
+`helix.log()` notes are available; failed evaluations still include stdout and
+stderr as fallback debug context.
 
 ---
 
--- a/src/helix/__init__.py
+++ b/src/helix/__init__.py
@@ -2,7 +2,11 @@
 
 from importlib.metadata import PackageNotFoundError, version as _pkg_version
 
+from helix.asi import log
+
 try:
     __version__ = _pkg_version("helix-evo")
 except PackageNotFoundError:  # not installed (e.g. source checkout without pip install -e .)
     __version__ = "0.0.0+unknown"
+
+__all__ = ["__version__", "log"]
--- a/src/helix/asi.py
+++ b/src/helix/asi.py
@@ -1,0 +1,82 @@
+"""Evaluator-facing arbitrary side information channel."""
+
+from __future__ import annotations
+
+import json
+import os
+from pathlib import Path
+from typing import Any
+
+
+HELIX_ASI_LOG_ENV = "HELIX_ASI_LOG"
+
+
+def log(*values: object, sep: str = " ", **fields: Any) -> None:
+    """Record evaluator notes for HELIX mutation prompts.
+
+    Evaluators can call ``from helix import log`` and then ``log(...)`` during
+    a HELIX-managed evaluation.  HELIX captures the notes through a per-
+    invocation file path in ``HELIX_ASI_LOG`` rather than through ordinary
+    stdout, keeping stdout free for machine protocols such as ``HELIX_RESULT``.
+
+    Outside a HELIX evaluator invocation this is a no-op, which lets evaluator
+    code keep the same imports in local debug runs.
+    """
+    path = os.environ.get(HELIX_ASI_LOG_ENV)
+    if not path:
+        return
+
+    record: dict[str, Any] = {}
+    if values:
+        record["message"] = sep.join(str(value) for value in values)
+    record.update(fields)
+    if not record:
+        return
+
+    try:
+        with Path(path).open("a", encoding="utf-8") as fh:
+            fh.write(json.dumps(record, sort_keys=True, default=str))
+            fh.write("\n")
+    except OSError:
+        return
+
+
+def read_text(raw: str) -> str:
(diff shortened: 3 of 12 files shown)


KE7:
@claude review

KE7:
# Code Review — PR #21: Add evaluator reserved metadata log channel

## Overview
Introduces a dedicated `helix.log()` API that lets evaluators emit human‑readable notes via a per‑invocation JSONL file (`HELIX_ASI_LOG`) instead of stdout. The executor wires the env var and captures the file (both host and sidecar/sandbox paths), while the mutator renders the notes as an `## Evaluator Notes` section and gates `stdout`/`stderr` to fallback‑only contexts (failures or absence of structured diagnostics). Stdout is now stripped of `HELIX_RESULT=` lines before mutation prompts. Coverage for executor env wiring, sandbox sequencing, log capture, prompt rendering, and failure fallback is included.

This is a clean, well‑scoped change. The contract (machine protocol on stdout, human notes via `helix.log`, raw streams reserved for failure/debug) is well motivated and consistently applied.

---

## Strengths
- **Good separation of concerns**: `reserved metadata.py` owns serialization (write/read/render/clear); executor owns lifecycle (path generation, env, cleanup); mutator owns prompt policy. Each part is unit‑testable in isolation.
- **Backward compatibility**: `log()` is a no‑op when `HELIX_ASI_LOG` is unset, so evaluators can keep the same imports in local debug runs.
- **Sandbox parity**: capturing the log file via an appended `cat` command in the sidecar command list mirrors the existing `extra_commands` plumbing rather than introducing a new sidecar API.
- **Tests cover the most important new branches**: env propagation, real subprocess capture, sandbox command ordering, machine‑protocol stripping, success suppression, and failure fallback.
- **Docs updated**: README explicitly distinguishes the machine protocol from `helix.log()` notes.

---

## Issues / Suggestions

### Correctness
- **`_returncode` leaks into `reserved metadata`** (`executor.py:_collect_asi`). It is filtered out by the mutator, but `EvalResult.reserved metadata` is consumed elsewhere (serialization, logs, downstream tooling). Recommend promoting it to a typed field on `EvalResult` (e.g. `evaluator_returncode: int | None`) rather than smuggling it as a stringified sentinel inside `reserved metadata`. The leading underscore convention is fragile across boundaries (JSON, hash keys, etc.).
- **`_evaluator_failed` always compares strings in practice.** `_collect_asi` writes `str(returncode)`, so the `0` int fallback in `not in (None, "0", 0)` is unreachable. Either drop the int branch, or — preferred — read the typed field suggested above.
- **Sandbox `cat` cwd assumption.** The capture command runs `cat <basename>` and depends on the sidecar evaluator's cwd being `/workspace` (where the file lives). This appears consistent with the rest of the sandbox path, but it is implicit. Consider `cat /workspace/<name>` for robustness, or assert the cwd contract in a comment near the `sh -c` line.
- **No sandbox‑side cleanup.** The non‑sandbox path uses `try/finally` + `clear_asi_log`, but the sandbox path leaves `.helix_asi_log_<uuid>.jsonl` in the worktree. Probably fine because worktrees are GC'd, but worth a comment so future readers don't assume the cleanup is symmetric. If worktrees can be reused, this becomes a real bug.
- **JSONL append is not atomic across writers.** `open(path, 'a')` + two `write()` calls (record + newline) is not multi‑process safe and not even multi‑threaded safe for records larger than `PIPE_BUF`. If an evaluator uses threads/multiprocessing, log lines can interleave. Either (a) write `payload + "\n"` in a single `write()` call, or (b) document that `helix.log()` is single‑writer.
- **Empty `log()` calls.** `log()` with neither values nor fields silently no‑ops. That's defensible, but a `log("")` call also no‑ops because `record["message"] = ""` is falsy under the early‑return — wait, actually `record` would be `{"message": ""}` which is truthy as a dict, so it would write. Worth confirming the intended semantics: do you want `log("")` to record an empty message?

### Code quality
- **`_collect_asi` docstring is stale.** It now adds `"log"` and `"_returncode"`, but the docstring's `Args` section omits `log`/`returncode` and only mentions `extra_0/extra_1`. Update to match the new keyword‑only params.
- **Extra commands also receive `HELIX_ASI_LOG`.** In non‑sandbox mode, every `extra_cmd` runs with the same env, so extras can also append notes to the main file. That's likely intentional but undocumented; a one‑line comment would help.
- **Mutator template chain** `{diagnostics_section}{evaluator_notes_section}{evaluator_output_section}{extra_asi_section}` relies on each helper to emit a trailing blank line when non‑empty and `""` otherwise. It works, but the invariant is implicit. A small comment on each `_render_*` helper documenting "returns either '' or text terminated by `\n\n`" would prevent future regressions.
- **`read_text` rendering** uses `f"{key}: {value}"` which will produce ugly output for nested dicts/lists logged via `**fields`. Consider `json.dumps(value, default=str)` for non‑scalar values.

### Tests
- Good coverage on new code paths. A few additions worth considering:
  - A regression test that asserts `_returncode` does **not** appear verbatim in mutation prompts (today only implicit via filter list).
  - A test that `helix.log()` outside a HELIX run (no env var) is a no‑op and does not raise.
  - A test for the `OSError` swallow in `log()` — currently silently dropping write errors is intentional but untested.
  - A sandbox test asserting the trailing capture command's `cat` output is parsed via `read_asi_log_text` (currently asserted only structurally on the command list).

### Security
- `helix_log_name` is `uuid4().hex`, which is filesystem‑safe; `shlex.quote` is applied for the sandbox shell command — both correct.
- Reading the file uses `Path.read_text` with `FileNotFoundError` swallowed. Other errors (e.g. permission, decode) will propagate; that's reasonable as it surfaces real misconfiguration.
- Stripping `HELIX_RESULT=` from the prompt is a nice defense against prompt‑injection of the machine protocol via stdout, but note evaluators can still inject misleading content via `helix.log()` itself. Probably out of scope for this PR.

### Performance
- Negligible overhead: one extra small file write per `log()` call and one extra `cat` invocation per evaluation in sandbox mode. Not a concern.

---

## Risk Assessment
- **Low** for the host (non‑sandbox) path: `try/finally` cleanup, env injection is local, behavior is gated on env presence.
- **Low–medium** for the sandbox path: relies on implicit cwd for the capture `cat`, and on worktree disposal for cleanup. Worth tightening before this becomes a hidden assumption.
- Behavioral change for users: prompts no longer always include stdout/stderr on success. Anyone who was implicitly relying on stdout reaching the mutator (without `helix.log()`) will see a behavior change. This is documented in the README — good — but consider calling it out in a release note / CHANGELOG entry.

---

## Recommendation
**Approve with minor changes.** The design is sound and the test plan is reasonable. Before merge, I'd ask for:
1. Promote `_returncode` to a typed field on `EvalResult` (or at minimum, document the underscore‑prefixed convention).
2. Update the `_collect_asi` docstring.
3. Add a brief comment about the sandbox cwd assumption and (lack of) cleanup symmetry.
4. Single‑`write()` JSONL emission, or an explicit single‑writer note.

Nice work — this is a clean improvement to the evaluator ↔ mutator contract.

PR #23, opened 2026-05-10
Fix protected directory evaluator manifests

KE7:
## Summary
- Recursively include regular files under protected evaluator directories in `.helix/evaluator_manifest.json` using stable repo-relative manifest keys.
- Reuse the protected-directory copy ignore patterns for manifest traversal so cache/tool artifacts such as `__pycache__`, `*.pyc`, `.pytest_cache`, `.mypy_cache`, and `.ruff_cache` are not hashed.
- Preserve existing protected-file hashing behavior and add focused evaluator-integrity coverage for directory recursion and ignored cache files.

## Root Cause
`_build_evaluator_integrity_manifest` collected protected paths correctly, but treated every protected path as a single regular file. Protected directories copied into candidate worktrees were skipped from the manifest, which produced warnings and left their contents out of tamper detection.

## Validation
- `uv run pytest tests/unit/test_evaluator_integrity.py -q`
- `uv run ruff check src/helix/evolution.py tests/unit/test_evaluator_integrity.py`
- `timeout 120s uv run mypy --strict src/helix/`

### DIFF
--- a/src/helix/evolution.py
+++ b/src/helix/evolution.py
@@ -74,6 +74,10 @@ from helix.worktree import (
 
 logger = logging.getLogger(__name__)
 
+# NOTE: ``shutil.ignore_patterns`` uses ``fnmatch`` *basename* matching, not
+# gitignore semantics.  Add basename globs here only -- patterns like
+# ``**/build/`` or ``cache/**`` will not match.  If we ever need gitignore
+# semantics, switch to ``pathspec`` rather than extending this tuple.
 _PROTECTED_DIRECTORY_IGNORE_PATTERNS = (
     ".git",
     "__pycache__",
@@ -81,6 +85,10 @@ _PROTECTED_DIRECTORY_IGNORE_PATTERNS = (
     ".pytest_cache",
     ".mypy_cache",
     ".ruff_cache",
+)
+# Reusable matcher; the patterns are constant so we build it once.
+_PROTECTED_DIRECTORY_IGNORE_MATCHER = shutil.ignore_patterns(
+    *_PROTECTED_DIRECTORY_IGNORE_PATTERNS
 )
 
 
@@ -263,7 +271,7 @@ def _copy_protected_path(source: Path, destination: Path) -> None:
         shutil.copytree(
             source,
             destination,
-            ignore=shutil.ignore_patterns(*_PROTECTED_DIRECTORY_IGNORE_PATTERNS),
+            ignore=_PROTECTED_DIRECTORY_IGNORE_MATCHER,
         )
     elif source.is_symlink():
         os.symlink(os.readlink(source), destination)
@@ -296,26 +304,41 @@ def _refresh_and_snapshot_protected_evaluator_files(
 
 
 def _sha256_file(path: Path) -> str:
-    return hashlib.sha256(path.read_bytes()).hexdigest()
+    """Streamed SHA-256 of ``path``.
+
+    Uses ``hashlib.file_digest`` (Python 3.11+) so large protected files
+    (e.g., evaluator dataset ``.jsonl`` files) are not buffered in memory.
+    """
+    with path.open("rb") as f:
+        return hashlib.file_digest(f, "sha256").hexdigest()
 
 
 def _iter_protected_manifest_files(
     source_path: Path,
     rel_path: str,
 ) -> list[tuple[str, Path]]:
-    """Return manifest entries under one protected file/directory."""
+    """Return manifest entries under one protected file/directory.
+
+    ``source_path`` is expected to already be a real path (callers resolve
+    symlinks before calling).  ``os.walk`` is used with ``followlinks=False``
+    (the default) so symlinks inside the directory are not traversed.
+    """
     if source_path.is_file():
         return [(rel_path, source_path)]
-    if not source_path.is_dir() or source_path.is_symlink():
+    if not source_path.is_dir():
         return []
 
-    ignore = shutil.ignore_patterns(*_PROTECTED_DIRECTORY_IGNORE_PATTERNS)
     entries: list[tuple[str, Path]] = []
     for dirpath_str, dirnames, filenames in os.walk(source_path):
-        ignored = ignore(dirpath_str, [*dirnames, *filenames])
+        ignored = _PROTECTED_DIRECTORY_IGNORE_MATCHER(
+            dirpath_str, [*dirnames, *filenames]
+        )
         dirnames[:] = sorted(name for name in dirnames if name not in ignored)
         for filename in sorted(name for name in filenames if name not in ignored):
             file_path = Path(dirpath_str) / filename
+            # Defensive against TOCTOU: ``os.walk`` already separated files
+            # from directories, but a concurrent unlink/rename could leave a
+            # stale name behind.
             if not file_path.is_file():
                 continue
             entry_path = Path(rel_path) / file_path.relative_to(source_path)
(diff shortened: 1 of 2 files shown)


KE7:
## Code Review — PR #23: Fix protected directory evaluator manifests

### Overview
Fixes a real bug: `_build_evaluator_integrity_manifest` previously treated every protected path as a regular file, so protected *directories* were skipped from the manifest with a warning and their contents were never tamper-checked. The PR now recursively enumerates regular files under protected directories using shared ignore patterns, preserving file-level behavior. The diff is small, well-scoped, and includes a focused test for directory recursion plus cache-file ignoring.

Overall this is a good fix and I'd land it after addressing a few items below.

### Strengths
- Correct root cause and minimal surgical fix.
- Good refactor: `_PROTECTED_DIRECTORY_IGNORE_PATTERNS` is now shared between `_copy_protected_path` and the manifest walker, eliminating drift.
- Deterministic output via `sorted(dirnames)` / `sorted(filenames)` per directory.
- `os.walk` is used with default `followlinks=False`, avoiding symlink-loop and escape risks.
- New test exercises both inclusion of nested files and exclusion of `__pycache__` / `.pytest_cache`, then verifies that tamper detection fires on the dataset file but not on changed cache content.

### Issues / Suggestions

1. **Dead symlink check on the call path.** In `_build_evaluator_integrity_manifest`, `source_path = (baseline_root / rel_path).resolve()` follows symlinks, so the inner `source_path.is_symlink()` guard inside `_iter_protected_manifest_files` is effectively unreachable from this caller. Either:
   - Drop the symlink branch (the `is_dir()` guard already covers the negative case), or
   - Move the symlink decision to `_build_evaluator_integrity_manifest` *before* `.resolve()` if you want to refuse symlinked top-level protected paths consistently with `_copy_protected_path`'s symlink handling.

2. **`_sha256_file` reads whole files into memory.** `path.read_bytes()` is fine for small files, but this PR meaningfully amplifies the blast radius: protected directories may now contain large datasets (e.g., `.jsonl`). Switch to streaming:
   ```python
   def _sha256_file(path: Path) -> str:
       with path.open("rb") as f:
           return hashlib.file_digest(f, "sha256").hexdigest()  # py3.11+
   ```
   or a chunked `while chunk := f.read(1 << 20)` loop. Pre-existing, but newly relevant.

3. **Tamper detection only catches modifications/deletions, not additions.** `_detect_evaluator_tamper` iterates `manifest.items()` and never walks the candidate worktree. After this PR a candidate could add a brand-new file under a protected directory (e.g., `datasets/leak.jsonl`) and it would not be flagged because it isn't in the baseline manifest. If the threat model includes injection of new evaluator inputs/scripts into a protected directory, the detector also needs to walk the candidate's protected dirs (using the same ignore patterns) and flag unknown files. Worth documenting this limitation explicitly even if you don't change behavior here.

4. **Per-file `OSError` is swallowed.** A failure to hash a single protected file logs an exception but proceeds to write a manifest *without* that entry. On resume, `_detect_evaluator_tamper` will then see the file as "present in worktree, absent in manifest"; i.e., not a violation. Consider failing closed (re-raise / abort manifest construction) so an unreadable protected file can't silently slip past tamper checks.

5. **Silent zero-entry directories.** If a protected directory exists but every file in it matches the ignore list, the manifest gets no entries for that path with no warning. Consider a `logger.warning` when a protected directory contributes 0 entries; useful signal for misconfigurations like `protected_files: ["splits"]` against a `splits/` that only contains caches.

6. **Ignore semantics are name-only fnmatch.** `shutil.ignore_patterns` matches against the basename via `fnmatch`, not gitignore semantics, so patterns like `**/build/` or `cache/**` won't behave as users might expect if the list grows. Fine for the current tuple; just worth a brief comment near `_PROTECTED_DIRECTORY_IGNORE_PATTERNS` so future contributors don't add gitignore-style globs by mistake.

7. **Minor: hoist the matcher.** `shutil.ignore_patterns(*_PROTECTED_DIRECTORY_IGNORE_PATTERNS)` is rebuilt on every call; the patterns are constant. You can construct it once at module scope (or memoize); negligible perf, slightly cleaner.

8. **Inner `is_file()` recheck is purely defensive.** `os.walk` already separates files from directories, so the `if not file_path.is_file(): continue` only catches a TOCTOU race. OK to keep, but a one-line comment would help future readers.

9. **Unrelated test rename mixed in.** The private example-id filename rename in `test_refreshes_protected_files_and_directories_from_current_root` and `test_copy_protected_path_noops_when_source_and_destination_match` is unrelated to the manifest fix. Not blocking, but ideally split into its own commit / PR for cleaner history.

10. **Manifest determinism on disk.** The dict is built in walk order, which is deterministic *given* a fixed filesystem, but file-system enumeration order can vary across platforms. Confirm that `_write_evaluator_integrity_manifest` writes with `sort_keys=True` (or sort here) so the on-disk JSON is stable for diffing/auditing across machines.

### Performance
- Hashing now scales with protected-directory size. Item (2) above is the main concern; item (3) would add a second walk if implemented. Both are tractable.

### Security
- `os.walk(followlinks=False)` is the right default: no symlink-escape risk.
- However, `_sha256_file` follows symlinks via `read_bytes()`, so a symlink *inside* a protected directory will be hashed by following its target. Likely acceptable for a tamper-detection use case, but worth being explicit about.
- Item (3) is the meaningful security gap: the manifest is allow-list-by-baseline, not a content-of-directory invariant. New files under protected dirs are invisible to the detector.

### Test coverage
- New test is well-targeted. Two extensions worth considering:
  - A case where a protected directory is empty / fully ignored (asserts behavior chosen for item 5).
  - A case asserting that a *new* file added to a protected directory is detected (or, until item 3 is addressed, asserting current behavior so the gap is documented in tests).

### Verdict
LGTM modulo items (1)-(3), which I'd treat as the must-fix set:
- (1) drop or relocate the dead symlink guard,
- (2) stream `_sha256_file`,
- (3) decide explicitly whether new files inside protected dirs should be flagged, and document/enforce that decision.

The remaining points are polish.

KE7:
## Review follow-up — pushed `20bffbb`

Addressed all 10 items from the review. Summary:

**Must-fix**
1. **Dead symlink guard** — removed the unreachable `source_path.is_symlink()` branch inside `_iter_protected_manifest_files`; added a docstring noting that callers `.resolve()` first and that `os.walk` runs with `followlinks=False`.
2. **Streaming hash** — `_sha256_file` now uses `hashlib.file_digest` (Py 3.11+) so large protected dataset files are no longer buffered in memory.
3. **New-file detection** — `_detect_evaluator_tamper` now optionally accepts `config` and `project_root`; when supplied, it walks the candidate worktree under each protected directory (using the shared ignore matcher) and flags any file that isn't in the frozen manifest. Both call sites in `_run_evolution_impl` now pass them. Legacy two-arg calls still work (only modifications/deletions reported), preserving the existing test contract.

**Polish**
4. Warn when a protected directory contributes zero manifest entries (catches misconfigs like `protected_files: ["splits"]` against a `splits/` that only contains caches).
5. **Fail closed on unreadable files** — `_build_evaluator_integrity_manifest` now raises `HelixError` instead of `logger.exception`-and-continue, so an unreadable protected file can't silently slip past tamper detection on resume.
6. Added a comment near `_PROTECTED_DIRECTORY_IGNORE_PATTERNS` documenting that `shutil.ignore_patterns` is fnmatch-basename-only (not gitignore), so future contributors don't add patterns that would silently no-op.
7. Hoisted the matcher to module scope (`_PROTECTED_DIRECTORY_IGNORE_MATCHER`); both `_copy_protected_path` and the manifest walker now reuse it.
8. Added a comment explaining the defensive `is_file()` recheck inside `_iter_protected_manifest_files` (TOCTOU guard, not a logic error).
9. Test rename left as-is (kept per request).
10. On-disk manifest sorting was already correct (`json.dumps(..., sort_keys=True)`); no change needed.

**New tests**
- `test_detects_new_files_added_under_protected_directory` — verifies the legacy two-arg contract still holds *and* that the four-arg form catches injected files while still ignoring cache files.
- `test_build_manifest_warns_for_directory_with_only_ignored_files`.
- `test_build_manifest_fails_closed_on_unreadable_protected_file`.

**Validation**
- `uv run pytest tests/unit -q` — **748 passed**
- `uv run ruff check src/helix/evolution.py tests/unit/test_evaluator_integrity.py` — clean
- `uv run mypy --strict src/helix/` — clean (28 files)

PR #25, opened 2026-05-10
Validate agent.effort against active backend

KE7:
## Summary
- Adds per-backend metadata for ``agent.effort`` in ``helix.backends`` (``EFFORT_AWARE_BACKENDS`` + ``EFFORT_VALID_VALUES``).
- Wires a config-load-time validator in ``HelixConfig.model_post_init`` that warns on (a) unknown effort values for backends that DO accept the field, and (b) any non-None effort on backends that silently ignore the field.
- Tightens the ``helix evolve --effort`` help text so the "when supported" caveat actually lists the supported backends and the per-backend mapping.

``helix resume`` deliberately does NOT expose ``--effort``: switching reasoning effort mid-run would mix candidates produced under different sampling regimes, breaking the comparability assumption that frontier acceptance and upstream reference-parity accounting depend on. Effort is a run-level setting only.

## Why
``agent.effort`` was being silently dropped on ``codex`` / ``cursor`` / ``gemini`` (no CLI equivalent) and unknown values on ``claude`` / ``opencode`` only showed up as opaque subprocess errors mid-run. Surfacing both at config load means users find out before spending an evolution budget on a setting that isn't taking effect.

Both checks are ``UserWarning``s rather than hard errors — users may legitimately need to forward a value the registry hasn't been updated for, and the backend CLI remains the source of truth for what's actually accepted. Same stance as the existing ``[evolution]`` upstream reference-parity warnings.

## Per-backend mapping (today)
| Backend | ``effort`` handling |
|---|---|
| ``claude`` | ``--effort <value>`` (allowlist: ``low|medium|high``) |
| ``opencode`` | ``--variant <value>`` (unrestricted; variant strings are model-specific) |
| ``codex`` | ignored — warning emitted |
| ``cursor`` | ignored — warning emitted |
| ``gemini`` | ignored — warning emitted |

## Tests
Nine cases under ``tests/unit/test_config.py::TestAgentEffortValidation``:
- ``effort=None`` -> silent.
- Each ignoring backend -> warning that names the backend.
- ``claude`` with each known value -> silent.
- ``claude`` with an unknown value -> warning that names the typo.
- ``opencode`` with an arbitrary variant string -> silent (unrestricted backend).
- "warning is a warning, not an error" — asserted via ``pytest.warns(UserWarning, match=...)`` so both invariants (constructs successfully + warning emitted) are explicit.
- **Registry-completeness invariant** (``test_every_backend_has_effort_metadata``): iterates over ``BACKENDS`` and asserts each entry produces the correct validator signal under a sentinel value. Catches drift the moment a new backend is added without updating ``EFFORT_AWARE_BACKENDS`` / ``EFFORT_VALID_VALUES``.

## Validation
- ``pytest`` -- **768 passed** (was 759 baseline; +9 new tests).
- ``ruff check src tests`` -- clean.
- Repo-wide ``ruff check .`` still blocked by the pre-existing ``F841`` in ``examples/web_researcher/evaluate.py`` -- same caveat as PR #18.

## Follow-up to
The "do we need to validate ``effort``?" thread on PR #18.

### DIFF
--- a/src/helix/backends.py
+++ b/src/helix/backends.py
@@ -28,9 +28,14 @@ BACKENDS: tuple[BackendName, ...] = ("claude", "codex", "cursor", "gemini", "ope
 EFFORT_AWARE_BACKENDS: frozenset[BackendName] = frozenset({"claude", "opencode"})
 """Backends that propagate ``agent.effort`` to their underlying CLI."""
 
-# ``None`` here means "any non-empty string accepted" — we still let a
-# strange value through and let the backend CLI decide, but a known set of
-# values lets HELIX warn early on obvious typos like ``effrot = "high"``.
+# ``None`` here means "any string accepted" — we still let strange values
+# through and let the backend CLI decide, but a known set of values lets
+# HELIX warn early on obvious typos like ``effrot = "high"``.
+#
+# NOTE: this allowlist is best-effort and may lag the upstream CLI.  When
+# a backend ships a new tier (e.g. Anthropic's ``"minimal"`` on some
+# surfaces), users will see a non-fatal "not a recognized value" warning
+# until this map is updated; the value still passes through to the CLI.
 EFFORT_VALID_VALUES: dict[BackendName, frozenset[str] | None] = {
     "claude": frozenset({"low", "medium", "high"}),
     "opencode": None,  # variant strings are model-specific; opencode validates them.
--- a/src/helix/backends.py
+++ b/src/helix/backends.py
@@ -9,6 +9,33 @@ BackendName: TypeAlias = Literal["claude", "codex", "cursor", "gemini", "opencod
 
 BACKENDS: tuple[BackendName, ...] = ("claude", "codex", "cursor", "gemini", "opencode")
 
+
+# ---------------------------------------------------------------------------
+# AgentConfig.effort metadata
+# ---------------------------------------------------------------------------
+#
+# ``agent.effort`` is the user-facing knob for "reasoning level / thinking
+# budget" on backends that expose one.  The string value is forwarded to a
+# backend-native CLI flag in ``helix.mutator``:
+#
+#   - ``claude``:    ``--effort <value>``   (Claude Code thinking budget)
+#   - ``opencode``:  ``--variant <value>``  (model variant selector)
+#   - others:        silently ignored (no equivalent CLI surface)
+#
+# The two registries below let the config layer fail fast on bad combinations
+# without hard-coding backend knowledge into ``HelixConfig``.
+
+EFFORT_AWARE_BACKENDS: frozenset[BackendName] = frozenset({"claude", "opencode"})
+"""Backends that propagate ``agent.effort`` to their underlying CLI."""
+
+# ``None`` here means "any non-empty string accepted" — we still let a
+# strange value through and let the backend CLI decide, but a known set of
+# values lets HELIX warn early on obvious typos like ``effrot = "high"``.
+EFFORT_VALID_VALUES: dict[BackendName, frozenset[str] | None] = {
+    "claude": frozenset({"low", "medium", "high"}),
+    "opencode": None,  # variant strings are model-specific; opencode validates them.
+}
+
 BACKEND_DISPLAY_NAMES: dict[str, str] = {
     "claude": "Claude Code",
     "codex": "Codex CLI",
--- a/src/helix/cli.py
+++ b/src/helix/cli.py
@@ -913,21 +913,14 @@ def log_cmd(project_dir: Path | None) -> None:
     type=click.Path(exists=True, file_okay=False, path_type=Path),
     help="Project root directory (defaults to current working directory).",
 )
-@click.option(
-    "--effort",
-    default=None,
-    help=(
-        "Override backend reasoning-effort level for the resumed run. "
-        "Forwarded as --effort to the 'claude' backend (low|medium|high) "
-        "and as --variant to the 'opencode' backend; ignored by "
-        "codex/cursor/gemini (a warning is emitted when set on a backend "
-        "that does not support it)."
-    ),
-)
-def resume(
-    config_path: str, project_dir: Path | None, effort: str | None
-) -> None:
+def resume(config_path: str, project_dir: Path | None) -> None:
     """Resume a previously started evolution run."""
+    # NOTE: ``resume`` deliberately does NOT expose ``--effort`` (or other
+    # agent-config overrides). Switching reasoning effort mid-run would
+    # mix candidates produced under different sampling regimes, breaking
+    # the comparability assumption that frontier acceptance and GEPA-parity
+    # accounting depend on. Effort is set once for the whole run via
+    # ``helix evolve --effort`` or the ``[agent]`` section of helix.toml.
     from helix.evolution import run_evolution
 
     project_root = project_dir if project_dir is not None else Path.cwd()
@@ -966,12 +959,6 @@ def resume(
     except FileNotFoundError:
         print_error(f"Config file not found: {cfg_file}")
         raise SystemExit(1)
-
-    # Apply CLI overrides (parity with ``helix evolve``).
-    if effort is not None:
-        config = config.model_copy(
-            update={"agent": config.agent.model_copy(update={"effort": effort})}
-        )
 
     print_info(f"Resuming from generation {state.generation if state else 0}…")
     try:
--- a/src/helix/cli.py
+++ b/src/helix/cli.py
@@ -913,7 +913,20 @@ def log_cmd(project_dir: Path | None) -> None:
     type=click.Path(exists=True, file_okay=False, path_type=Path),
     help="Project root directory (defaults to current working directory).",
 )
-def resume(config_path: str, project_dir: Path | None) -> None:
+@click.option(
+    "--effort",
+    default=None,
+    help=(
+        "Override backend reasoning-effort level for the resumed run. "
+        "Forwarded as --effort to the 'claude' backend (low|medium|high) "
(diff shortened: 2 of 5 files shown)


KE7:
## Code Review — PR #25 (Validate `agent.effort` against active backend)

### Overview
Adds a config-load-time `UserWarning` for two failure modes of `agent.effort`:
1. The selected backend silently ignores the field (`codex` / `cursor` / `gemini`).
2. The selected backend is effort-aware but the value isn't in the known allowlist (e.g. `claude` with `effort=\"extreme\"`).

The truth table lives in `helix.backends` (`EFFORT_AWARE_BACKENDS` + `EFFORT_VALID_VALUES`) so `HelixConfig` stays free of per-backend `if`s. CLI help text for `helix evolve --effort` is also tightened to enumerate the supported backends and the per-backend flag mapping.

Overall this is a clean, well-scoped change with good test coverage. Comments below are mostly nits / follow-ups.

---

### Correctness

- ✅ Validator short-circuits on `effort is None`, so the default path is silent (covered by `test_effort_unset_is_silent`).
- ✅ `EFFORT_VALID_VALUES.get(backend)` returning `None` is treated as \"unrestricted\" — matches the inline docstring and `opencode`'s actual behavior (`--variant` accepts arbitrary model-specific strings).
- ✅ `mutator.py` already keys off `config.effort` truthiness at lines 689 and 748, so the registries in `backends.py` accurately reflect runtime behavior.
- ⚠️ **`stacklevel=3` is unusual here.** Other warnings in the codebase (`budget.py`, `state.py`) use `stacklevel=2`. From `_validate_agent_effort` the call stack at `warnings.warn` time is roughly: `warn` → `_validate_agent_effort` → `model_post_init` → pydantic internals → user `HelixConfig(...)` call. `stacklevel=3` lands inside pydantic's `__init__`, not user code. If the goal is to attribute the warning to the user's `HelixConfig(...)` call site, you'd need a higher stacklevel (and even that is brittle across pydantic versions). Easiest fix: drop to `stacklevel=2` for consistency with the rest of the repo, or accept that pydantic post-init warnings aren't reliably re-attributable.
- ⚠️ **PR description vs. diff drift.** The summary says \"Tightens the `helix evolve --effort` / `helix resume --effort` help text,\" but `helix resume` doesn't actually expose an `--effort` flag (verified at `cli.py:897-916`). Either the description should be corrected or `resume` should grow the same override flag for parity. Not blocking.

### Style / Conventions

- 👍 Putting the metadata next to `BACKENDS` in `backends.py` keeps a single source of truth for backend-shape facts.
- 👍 Frozensets for the allowlists — immutable, hashable, and signal intent.
- Nit: the local `import warnings as _warnings` inside `_validate_agent_effort` is justified in the comment as \"keep config import-light,\" but `warnings` is a stdlib module Python imports at startup anyway — the savings are essentially zero. Consider lifting to the module-level imports for readability.
- Nit: `f\"backend does not propagate an effort/reasoning level — the value will be silently ignored.\"` mixes an em-dash with a period inside an f-string that's then concatenated with a non-f-string (`\"value will be silently ignored. ...\"`). Functionally fine, but the message reads slightly disjointed; a single triple-quoted message would be cleaner.
- The comment in `backends.py` saying `\"None ... means 'any non-empty string accepted'\"` is slightly stronger than the code enforces — the code accepts *any* string, including empty strings, when `valid is None`. Either tighten the doc to \"any string\" or add an empty-string guard. (Pydantic doesn't reject `effort=\"\"` today either.)

### Test Coverage

- ✅ Eight new cases under `TestAgentEffortValidation` cover: unset, each ignoring backend, each valid claude value, claude typo, opencode unrestricted, and the warning-not-error invariant.
- ✅ `recwarn` fixture is used cleanly and assertions filter on `UserWarning` so unrelated warnings don't pollute the result.
- 💡 **Suggested additional test (registry-completeness invariant):** assert that for every name in `BACKENDS`, the backend is either in `EFFORT_AWARE_BACKENDS` *or* will trigger the \"does not propagate\" warning. This guards against a new backend being added to `BACKENDS` without anyone updating the effort registries — the validator would silently start emitting the wrong message (or no message) for the new backend.
  ```python
  def test_every_backend_has_effort_metadata(self, recwarn):
      from helix.backends import BACKENDS, EFFORT_AWARE_BACKENDS
      for backend in BACKENDS:
          assert (
              backend in EFFORT_AWARE_BACKENDS
              or backend not in EFFORT_AWARE_BACKENDS  # explicit ignore
          )
          # And the validator must produce *some* signal:
          ...
  ```
- 💡 `test_effort_warning_is_warning_not_error` has no explicit assertion — it relies on the constructor not raising. Consider wrapping in `pytest.warns(UserWarning)` to make the invariant (\"warning, not error, AND a warning is emitted\") explicit.
- 💡 No test pins the exact backends that ignore effort to a list. If `codex` later grows an effort flag and is added to `EFFORT_AWARE_BACKENDS`, `test_effort_warns_on_ignoring_backend_codex` will start failing — which is correct, but a parametrized test driven off `set(BACKENDS) - EFFORT_AWARE_BACKENDS` would be more maintenance-proof.

### Performance / Security

- No performance concerns — the validator runs once per config load, O(1).
- No security concerns — purely advisory, no privilege boundary crossed.

### Risks

- **Registry-drift risk** (the main one): `EFFORT_AWARE_BACKENDS` / `EFFORT_VALID_VALUES` and `mutator.py`'s effort-handling branches must stay in lockstep. The completeness test suggested above mitigates this for the \"new backend added\" case but not for \"existing backend's CLI changes its allowlist.\" A short comment in `mutator.py` near the two `if config.effort` blocks pointing at `backends.EFFORT_AWARE_BACKENDS` would help future maintainers.
- **Forward-compat allowlist**: Anthropic recently shipped `\"minimal\"` as a Claude effort tier on some surfaces. The current `frozenset({\"low\", \"medium\", \"high\"})` will warn on it. Since the warning is non-fatal and the value still passes through, this is acceptable, but worth an inline TODO so the next reader knows it's an allowlist that may lag the upstream CLI.

### Verdict
Looks good. The two things I'd most want to see before merge are:
1. A **registry-completeness test** (every `BACKENDS` entry is classified).
2. A decision on **`stacklevel`** (pick `2` for repo consistency, or document why `3`).

Everything else is nit-level.

KE7:
## Review feedback addressed in 4b75deb

Pushed a follow-up commit that addresses every issue raised in the review above. Summary of changes:

### config.py
- Lifted `import warnings` to module scope (the local-import \"keep config import-light\" rationale doesn't save anything for a stdlib module that Python pre-imports anyway).
- `stacklevel=3` → `stacklevel=2` for consistency with `helix.budget` and `helix.state`.
- Tightened warning message wording and de-fragmented the f-string / non-f-string concatenation into a single readable string per call.

### backends.py
- Doc fix: `EFFORT_VALID_VALUES` comment now says \"any string accepted\" (not \"any non-empty string\"), matching what the code actually enforces.
- Added a forward-compat note flagging that the allowlist may lag the upstream CLI (e.g. Anthropic's `\"minimal\"` tier on some surfaces).

### mutator.py
- Added cross-reference comments to both `effort`-handling branches (`claude --effort` and `opencode --variant`) pointing at `helix.backends.EFFORT_AWARE_BACKENDS` / `EFFORT_VALID_VALUES`, so future maintainers know the registry and CLI args must stay in sync.

### cli.py
- Added `helix resume --effort` for parity with `helix evolve --effort`. The original PR description claimed this had been tightened but `resume` had no such flag — now it does, with the same help text and override semantics.

### tests/unit/test_config.py
- `test_effort_warning_is_warning_not_error` now uses `pytest.warns(UserWarning, match=\"does not propagate\")` so both invariants (HelixConfig constructs successfully *and* a warning is emitted) are explicit, not implicit.
- New `test_every_backend_has_effort_metadata`: a registry-completeness invariant that iterates over `BACKENDS` and asserts each one produces the correct validator signal under a sentinel `effort` value:
  - ignoring backends (`codex` / `cursor` / `gemini`) → \"does not propagate\" warning
  - aware + restricted allowlist (`claude`) → \"not a recognized value\" warning
  - aware + unrestricted (`opencode`, `valid is None`) → silent (any string allowed)
  - This catches drift the moment a new backend is added to `BACKENDS` without `EFFORT_AWARE_BACKENDS` / `EFFORT_VALID_VALUES` being updated alongside it.

### Validation
- `uv run pytest` — **768 passed** (was 767; +1 new test from the registry-completeness invariant).
- `uv run ruff check src tests` — clean.
- `uv run helix resume --help` — confirms the new `--effort` flag is wired correctly.

Ready for re-review.

KE7:
## Reverted ``helix resume --effort`` in 8accb01

Reviewer pushback (correct): switching reasoning effort mid-run would mix candidates produced under different sampling regimes, breaking the comparability assumption that frontier acceptance and upstream reference-parity accounting depend on. Effort is a run-level setting; ``helix evolve --effort`` and the ``[agent]`` section of ``helix.toml`` are the only legitimate surfaces.

I added the flag in 4b75deb thinking the original PR description's claim of \"tightens ``helix resume --effort``\" implied the surface should exist for parity. The right fix was the opposite: leave ``resume`` alone and correct the PR description. Done in 8accb01:

- ``cli.py``: removed the ``--effort`` option from ``resume``; added an inline NOTE in the resume command body explaining the deliberate omission so a future contributor doesn't try to re-add it.
- PR description: dropped the ``helix resume --effort`` line; added an explicit \"resume deliberately does NOT expose --effort\" paragraph.

### Validation
- ``uv run pytest`` -- **768 passed** (unchanged, no test regression from the revert).
- ``uv run ruff check src tests`` -- clean.
- ``uv run helix resume --help`` -- confirms only ``--config`` and ``--dir`` remain on resume.

PR #26, opened 2026-05-10
Recover sandbox sync after userns remap

KE7:
## Summary
- Recover sandbox workspace readability before host-side sync-back when direct host ownership restoration is unavailable.
- Use a root helper container to relax workspace mode bits in user-namespace/rootless Docker environments.
- Reuse the same recovery path for temporary sandbox cleanup.

## Root Cause
Some Docker configurations remap container-owned files to host UIDs that the host process cannot read after the agent command exits. HELIX already skips direct host chown in those configurations, but without a fallback the subsequent host-side sync-back can preserve stale files or skip artifacts because the temporary workspace is inaccessible.

## Validation
- `uv run pytest tests/unit/test_sandbox.py -q`
- `uv run ruff check src/helix/sandbox.py tests/unit/test_sandbox.py`
- Manual Docker sync-back smoke test with a container-created file and restrictive directory permissions

### DIFF
--- a/src/helix/sandbox.py
+++ b/src/helix/sandbox.py
@@ -362,14 +362,17 @@ def _safe_rmtree(path: Path, *, docker_image: str | None = None) -> None:
     except FileNotFoundError:
         return
     except OSError as first_exc:
-        if docker_image and (host_owner := _host_owner()):
-            _docker_chown_workspace(path, docker_image, host_owner)
+        if docker_image:
+            if host_owner := _host_owner():
+                _docker_chown_workspace(path, docker_image, host_owner)
+            else:
+                _docker_relax_workspace_permissions(path, docker_image)
             try:
                 shutil.rmtree(path)
                 return
             except OSError as second_exc:
                 logger.warning(
-                    "failed to clean sandbox temp dir %s after chown: %s",
+                    "failed to clean sandbox temp dir %s after permission recovery: %s",
                     path,
                     second_exc,
                 )
@@ -490,6 +493,47 @@ def _docker_chown_workspace(workspace: Path, image: str, owner: str) -> None:
             "-c",
             'find /workspace -path /workspace/.git -prune -o -exec chown -h "$0" {} +',
             owner,
+        ],
+        check=False,
+        capture_output=True,
+        text=True,
+        env=_docker_host_env(),
+    )
+
+
+def _docker_relax_workspace_permissions(workspace: Path, image: str) -> None:
+    """Make a sandbox workspace readable/writable by the host after userns runs.
+
+    Rootless Docker and user-namespace remapping can make container-owned files
+    appear on the host as unmapped high UIDs. In that mode chowning to the host
+    UID from inside the container is not meaningful, but a root helper in the
+    same namespace can still relax mode bits on the bind mount before host-side
+    sync and cleanup.
+    """
+    subprocess.run(
+        [
+            "docker",
+            "run",
+            "--rm",
+            "--workdir",
+            "/workspace",
+            "--user",
+            "root",
+            "--network",
+            "none",
+            "--security-opt",
+            "no-new-privileges",
+            "-v",
+            f"{workspace}:/workspace:rw",
+            image,
+            "sh",
+            "-c",
+            (
+                "find /workspace -path /workspace/.git -prune -o "
+                "-type d -exec chmod 0777 {} +; "
+                "find /workspace -path /workspace/.git -prune -o "
+                "-type f -exec chmod 0666 {} +"
+            ),
         ],
         check=False,
         capture_output=True,
@@ -927,6 +971,8 @@ def run_sandboxed_commands(
         finally:
             if host_owner := _host_owner():
                 _docker_chown_workspace(workspace, docker_image, host_owner)
+            else:
+                _docker_relax_workspace_permissions(workspace, docker_image)
         if sync_back:
             _sync_back_workspace(
                 workspace,
--- a/src/helix/sandbox.py
+++ b/src/helix/sandbox.py
@@ -472,32 +472,67 @@ def _init_synthetic_git_repo(workspace: Path) -> None:
     )
 
 
-def _docker_chown_workspace(workspace: Path, image: str, owner: str) -> None:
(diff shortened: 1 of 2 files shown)


KE7:
## Code Review: Recover sandbox sync after userns remap

### Overview
Adds a `_docker_relax_workspace_permissions` helper that runs a root-in-container `chmod` pass over the bind-mounted workspace whenever direct host `chown` is unavailable (rootless Docker / user-namespace remap). The helper is wired into:
- the post-run `finally` block of `run_sandboxed_commands` (before host-side sync-back), and
- `_safe_rmtree`'s recovery branch.

The change cleanly mirrors the existing `_docker_chown_workspace` helper and unblocks sync-back/cleanup on hosts where `_host_owner()` returns `None`.

### Strengths
- Diagnosis is correct: under userns remap the container's root maps to the subuid that owns the bind-mounted files, so a root helper can in fact relax the mode bits even though chowning to a host UID would be meaningless.
- New helper consistently reuses `--user root`, `--network none`, `--security-opt no-new-privileges`, `_docker_host_env()`, and the `.git` prune pattern, matching the existing chown helper.
- Test coverage exercises the new path end-to-end: it confirms the relax helper is invoked after the agent container exits and that sync-back picks up container-modified files.
- Log-message tweak (`after permission recovery` instead of `after chown`) reflects the broadened recovery behavior.

### Issues / Suggestions

**1. `chmod 0666`/`0777` strips the executable bit on regular files (correctness).**
Many sandboxed flows produce executable scripts (e.g., generated test runners, hooks committed to the synthetic repo). Hard-coding mode `0666` on regular files will silently drop `+x` before sync-back, so executables created in the container land back on the host as non-executable. A safer canonical pattern is:

```sh
find /workspace -path /workspace/.git -prune -o -exec chmod a+rwX {} +
```

`a+rwX` (capital `X`) adds `r`/`w` to everything and adds `x` only to directories or files that already had `x` for any user. That keeps directories traversable, files writable, and preserves existing executable bits. It also collapses the two `find` invocations into one.

**2. Silent failure of the helper container.**
Like `_docker_chown_workspace`, the new helper uses `check=False` with `capture_output=True` but never inspects `returncode`/`stderr`. If the helper fails (image missing, daemon error, OOM, mount issue) the only signal is a downstream `sync_back`/`rmtree` error whose root cause is then hidden. Consider:

```python
result = subprocess.run(..., check=False, capture_output=True, text=True, ...)
if result.returncode != 0:
    logger.warning(
        "docker permission-relax helper failed for %s: rc=%s stderr=%s",
        workspace, result.returncode, result.stderr.strip(),
    )
```

Worth applying the same logging to `_docker_chown_workspace` while you're here — both are silent today.

**3. Security note: world-writable workspace on the host.**
After the helper runs, the temp workspace is `0777`/`0666` on the host. It's a per-run temp dir under HELIX's control, but if any host process (or a cohabiting user on a shared machine) can race the sync-back window, it could mutate workspace files between the helper exit and `_sync_back_workspace`. In practice the temp parent's mode usually mitigates this, but it's worth a one-line comment acknowledging the trade-off, or scoping mode to `0700`/`0600` for owner-only after the chmod (which would still be writable by the host user since the host now sees the files via the relaxed mode bits and matching UID context).

**4. `_safe_rmtree` redundancy.**
When `run_sandboxed_commands` already calls `_docker_relax_workspace_permissions` in its `finally`, the subsequent `_safe_rmtree(tmp_path, ...)` will normally succeed on the first `shutil.rmtree`. If it doesn't, `_safe_rmtree` invokes the helper a second time. That's fine functionally, just an extra ~1s container start in the failure path — no action needed, but worth noting.

**5. Test gaps.**
- No assertion that the relax helper is *not* invoked when `_host_owner()` returns a real value (i.e., that the new `else` branch is mutually exclusive).
- No coverage for the `_safe_rmtree` recovery branch via the relax path (the new test only covers `run_sandboxed_commands`).
- The ordering assertion is correct but a bit dense; an inline comment explaining "relax must run after the agent container and before host sync-back reads the workspace" would help future readers.

**6. Minor style.**
- The two `find … -exec chmod` invocations strung together with `;` in `sh -c` work, but per (1) you can collapse to a single invocation with `a+rwX`.
- The helper duplicates ~25 lines of `docker run` boilerplate from `_docker_chown_workspace`. Extracting a small `_docker_run_workspace_helper(workspace, image, sh_command)` would deduplicate and make the security-flag set easier to keep in sync across helpers.

### Performance
Adds one extra short-lived container launch per sandboxed run on rootless/userns hosts (and one more on `_safe_rmtree` failure). Negligible with a cached image; the only watch-out is that the helper image is the same `docker_image` used for the agent, which is already warm.

### Risk Summary
- Functional risk: **low** — change is additive and gated on `_host_owner() is None`.
- Behavioral risk: **moderate** for any workflow that relies on executable-bit preservation across sync-back (see issue 1). Recommend switching to `a+rwX` before merge.
- Observability risk: **moderate** — silent helper failures would manifest as confusing downstream sync errors; please add a returncode log.

### Recommendation
Approve with the `a+rwX` fix and a returncode log line; the rest are nice-to-haves.

KE7:
Pushed `2fca10d` addressing the review feedback:

- **`chmod a+rwX`** replaces the flat `0666`/`0777` pair so executable bits and directory-traversal bits survive sync-back; the two `find` invocations collapse into one.
- **Returncode/stderr logging**: extracted shared `docker run` boilerplate into `_run_workspace_helper`; both the chown and relax helpers now log on non-zero exit instead of silently masking failures as confusing downstream sync/cleanup errors.
- **Test coverage**:
  - Negative-path: `test_agent_sync_skips_relax_when_host_owner_available` confirms the chown branch and relax branch are mutually exclusive.
  - `_safe_rmtree` recovery: `test_safe_rmtree_uses_relax_helper_when_host_owner_missing` confirms the rootless rmtree retry path uses the relax helper, not chown.

`uv run pytest tests/unit/test_sandbox.py -q` → 31 passed; `uv run ruff check` → clean.

Skipped per scope: the world-writable workspace is still transient and the previous behavior — same security envelope as the existing chown helper, no functional change there.

PR #27, opened 2026-05-10
Propagate backend reasoning effort

KE7:
## Summary
- Propagate `agent.effort` to `codex exec` via `-c model_reasoning_effort=<value>`.
- Mark Codex as effort-aware in backend validation, including recognized Codex effort values.
- Update CLI help text and focused tests for Codex effort propagation and TOML escaping.

## Validation
- `uv run --extra dev pytest tests/unit/test_mutator.py tests/unit/test_config.py`
- `uv run --extra dev ruff check src/helix/backends.py src/helix/config.py src/helix/cli.py src/helix/mutator.py tests/unit/test_config.py tests/unit/test_mutator.py`

### DIFF
--- a/src/helix/backends.py
+++ b/src/helix/backends.py
@@ -19,13 +19,16 @@ BACKENDS: tuple[BackendName, ...] = ("claude", "codex", "cursor", "gemini", "ope
 # backend-native CLI flag in ``helix.mutator``:
 #
 #   - ``claude``:    ``--effort <value>``   (Claude Code thinking budget)
+#   - ``codex``:     ``-c model_reasoning_effort=<value>`` (Codex CLI config)
 #   - ``opencode``:  ``--variant <value>``  (model variant selector)
 #   - others:        silently ignored (no equivalent CLI surface)
 #
 # The two registries below let the config layer fail fast on bad combinations
 # without hard-coding backend knowledge into ``HelixConfig``.
 
-EFFORT_AWARE_BACKENDS: frozenset[BackendName] = frozenset({"claude", "opencode"})
+EFFORT_AWARE_BACKENDS: frozenset[BackendName] = frozenset(
+    {"claude", "codex", "opencode"}
+)
 """Backends that propagate ``agent.effort`` to their underlying CLI."""
 
 # ``None`` here means "any string accepted" — we still let strange values
@@ -38,6 +41,7 @@ EFFORT_AWARE_BACKENDS: frozenset[BackendName] = frozenset({"claude", "opencode"}
 # until this map is updated; the value still passes through to the CLI.
 EFFORT_VALID_VALUES: dict[BackendName, frozenset[str] | None] = {
     "claude": frozenset({"low", "medium", "high"}),
+    "codex": frozenset({"minimal", "low", "medium", "high", "xhigh"}),
     "opencode": None,  # variant strings are model-specific; opencode validates them.
 }
 
--- a/src/helix/cli.py
+++ b/src/helix/cli.py
@@ -599,9 +599,10 @@ def sandbox_logout(
     default=None,
     help=(
         "Override backend reasoning-effort level. Forwarded as --effort to "
-        "the 'claude' backend (low|medium|high) and as --variant to the "
-        "'opencode' backend; ignored by codex/cursor/gemini (a warning is "
-        "emitted when set on a backend that does not support it)."
+        "the 'claude' backend (low|medium|high), as model_reasoning_effort "
+        "to the 'codex' backend, and as --variant to the 'opencode' backend; "
+        "ignored by cursor/gemini (a warning is emitted when set on a backend "
+        "that does not support it)."
     ),
 )
 def evolve(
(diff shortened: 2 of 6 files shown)


KE7:
## Code Review — [codex] Propagate Codex reasoning effort

### Overview
This PR wires up `agent.effort` to Codex CLI via `-c model_reasoning_effort=<value>`, bringing it in line with the Claude (`--effort`) and OpenCode (`--variant`) backends. The change is well-scoped: it touches exactly the four files that need updating (`backends.py`, `mutator.py`, `cli.py`, `config.py`) and ships corresponding tests. Overall structure is clean and follows existing conventions closely.

---

### ✅ What's done well

- **Registry-first design is respected.** Adding codex to both `EFFORT_AWARE_BACKENDS` and `EFFORT_VALID_VALUES` keeps the config-validation layer in sync with the mutator, exactly as the comment at the top of `backends.py` prescribes.
- **JSON quoting for the value.** Using `json.dumps(config.effort)` to produce the quoted string (`"high"`) rather than naive f-string concatenation is the right move and protects against injection of characters like `"` or `\`.
- **Differentiated valid-value sets.** Codex's `{"minimal", "low", "medium", "high", "xhigh"}` is correctly separated from Claude's `{"low", "medium", "high"}` rather than sharing a constant.
- **Test coverage.** The PR replaces the old "warns when effort is set on codex" test with "accepts all valid values without warning," and adds a corresponding "warns on unknown value" test — symmetrical with the Claude tests. The TOML-escape test using `high"quoted` is a nice edge-case regression guard.
- **`test_effort_warning_is_warning_not_error`** is correctly migrated to use `gemini` as the "ignoring backend" so it still exercises the warning path.

---

### 🔶 Issues & suggestions

#### 1. Test name vs. implementation: "TOML escaped" ≠ `json.dumps`
```python
def test_codex_effort_config_value_is_toml_escaped(self, mocker):
```
The test name says "TOML escaped" but the implementation uses `json.dumps`, which produces **JSON** string literals. JSON and TOML basic-string quoting rules overlap for ASCII but diverge for edge cases (e.g., TOML allows `\uXXXX` only for code points ≥ U+0020, while JSON requires escaping U+0000–U+001F). For normal effort strings this is irrelevant in practice, but the mismatch in terminology will confuse future maintainers.

**Suggestion:** Rename the test to `test_codex_effort_config_value_is_json_quoted` (or add a brief inline comment explaining why `json.dumps` is being used here), and update the PR description's "TOML escaping" mention accordingly.

#### 2. No inline comment explaining the `-c key=value` convention
The other backends have explanatory comments above their effort blocks:
```python
# ``effort`` is the user-facing knob for reasoning-level / thinking
# budget; the registry of backends that honor it lives in …
```
The new codex block has no equivalent comment explaining why the `-c model_reasoning_effort=<value>` pattern is used. A one-liner referencing the Codex CLI docs or the `codex exec --help` output would help future readers understand the interface contract.

#### 3. `"xhigh"` tier — source of truth?
`"xhigh"` is listed as a valid effort value for codex. A brief comment or a link to the Codex CLI changelog/docs confirming this tier exists would prevent the allowlist from becoming stale silently. (The module-level comment already acknowledges the allowlist may lag, but a reference helps.)

#### 4. Empty-string effort is falsy — consistent but worth noting
```python
if config.effort:
    args.extend(["-c", f"model_reasoning_effort={json.dumps(config.effort)}"])
```
This is consistent with the `claude` and `opencode` branches, so it's not a regression, but an empty string `""` would silently be skipped. If `AgentConfig` validates that `effort` is either `None` or a non-empty string this is fine; just worth a quick check that the dataclass/pydantic model enforces that.

---

### 🔍 Minor nits

- **CLI help text** (cli.py) reads `"as model_reasoning_effort to the 'codex' backend"` — consider `"as -c model_reasoning_effort=<value> to the 'codex' backend"` to match the actual flag form and help users understand what they'd see in shell output.
- The config.py docstring update removes codex from the "silently ignored" list but the surrounding prose still says "two cases are surfaced" — that count remains correct (ignoring + bad value), so no action needed there; just confirming.

---

### Summary

| Aspect | Assessment |
|--------|------------|
| Correctness | ✅ Effort flag correctly placed before prompt instruction; JSON quoting is safe |
| Conventions | ✅ Follows existing backend-arg builder pattern exactly |
| Test coverage | ✅ Happy path, warning path, and escape edge case all covered |
| Terminology | ⚠️ "TOML escaped" vs `json.dumps` — rename the test |
| Documentation | ⚠️ Missing inline comment on the `-c` convention; CLI help could be more precise |
| Risk | 🟢 Low — change is additive and doesn't affect other backends |

The one actionable change before merge is the test renaming (issue #1). The rest are minor polish suggestions.

KE7:
Addressed both review issues in the latest commit (`1417723`):

1. **Test renamed** — `test_codex_effort_config_value_is_toml_escaped` → `test_codex_effort_config_value_is_json_quoted`, with an inline comment explaining why `json.dumps()` output is safe for TOML basic-string contexts.

2. **Inline comment added** in `mutator.py` above the `-c key=value` construction documenting the Codex CLI convention and the rationale for using `json.dumps`.

All tests still pass (`TestInvokeClaudeCode` · 19 passed, `TestAgentEffortValidation` · 10 passed).

PR #37, opened 2026-05-20
feat(cache): preserve per-(candidate, example) side_info across cache hits

KE7:
## Summary

Preserves per-example evaluator `side_info` across evaluation-cache hits so useful mutation feedback remains available on resume and repeated examples.

## Disclosure fixes

- The raw evaluator result and cache remain unchanged for parsing and cache fidelity. At mutation-prompt rendering, diagnostics are recursively sanitised to replace absolute filesystem paths while retaining keys and non-sensitive feedback.
- Configured environment values are redacted at the same final prompt boundary using the shared value-aware redactor. It preserves dictionary keys and non-sensitive values.
- Cache schema-mismatch warnings now report the version mismatch and retained diagnostic copy without exposing filesystem paths.

## Shared component

`src/helix/redaction.py` is cherry-picked from PR #47, which owns it; merge #47 first or this commit will appear twice in history. This PR uses that module unchanged for configured secret values. The local prompt-boundary sanitiser separately covers filesystem paths, which are provenance data rather than configured secret values.

## Test plan

- [x] `uv sync --extra dev`
- [x] `uv run pytest tests/unit -q` — 880 passed
- [x] `uv run mypy --strict src/helix/` — clean
- [x] Regression coverage: nested side-info paths do not reach the prompt; raw evaluator data remains unchanged; useful keys and non-sensitive values survive; configured values are masked; schema-mismatch warnings contain no absolute paths.

### DIFF
--- a/src/helix/batch_sampler.py
+++ b/src/helix/batch_sampler.py
@@ -1,8 +1,8 @@
 """HELIX minibatch sampler — GEPA parity.
 
-Line-for-line port of
-  gepa.strategies.batch_sampler.EpochShuffledBatchSampler
-(see /tmp/gepa_eval_spec.md §2).
+Port of ``gepa.strategies.batch_sampler.EpochShuffledBatchSampler``
+(see ``src/gepa/strategies/batch_sampler.py`` in the public GEPA repo,
+``github.com/gepa-ai/gepa``).
 
 Also provides :class:`StratifiedBatchSampler`, which is a HELIX extension
 over EpochShuffledBatchSampler that guarantees each minibatch of size K
--- a/src/helix/config.py
+++ b/src/helix/config.py
@@ -344,7 +344,7 @@ class EvolutionConfig(BaseModel):
             "Max parallel eval workers — bounds both the parent-eval and "
             "mutation ThreadPools in the num_parallel_proposals pipeline. "
             "GEPA parity: EngineConfig.max_workers "
-            "(/tmp/gepa-official/src/gepa/optimize_anything.py:485, "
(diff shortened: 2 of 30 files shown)


KE7:
Closing: superseded, and the conflict is the evidence.

This PR's headline commit `8c60a71 feat(cache): preserve per-(candidate, example) side_info across cache hits` (2026-05-20) was superseded the next day by `321d01a Cache per-example side info` on `main` (2026-05-21), which is a different implementation of the same feature. `git merge-base --is-ancestor 321d01a feat/cache-side-info-parity` is false — the two never met, which is exactly why `eval_cache.py` conflicts.

The branch is 11 commits behind main and conflicting across `evolution.py`, `mutator.py`, `config.py`, `eval_cache.py`, `batch_sampler.py` and six test files. GitHub cannot build a merge ref for a conflicting PR, which is why no CI has run here since 2026-05-20.

Disposition of the remaining commits:
- `9a4af63` (shared diagnostic redactor) — owned by #47, already merged there.
- `712c9a3` (redact diagnostic paths at prompt boundary) — depends on `helix.redaction`; the underlying Diagnostics leak is already fixed on #47.
- `0423fec` (avoid paths in cache schema warning) — fixes a warning introduced by `07a4175` in this PR, so it dies with it. The equivalent defect on main (`state.py:302-314`) is being tracked separately.
- `07a4175` (invalidate pre-extension `eval_cache.pkl`) — guards a cache format this PR would have introduced; main's format came from `321d01a` instead.
- `defb748` — already on main.

Branch retained.

PR #43, opened 2026-07-24
ci: automate pinned mutation runner image publication

KE7:
Rebased onto a stack of three independently-mergeable PRs. **The backup of the previous tip (`10f8c5f`) is `backup/runner-refresh-pre-split-10f8c5f`, pushed.**

**This PR: 10 files, +2,951 / −225.**

```
git diff --shortstat origin/stack/runner-images-base...origin/agent/helix-runner-refresh
```

Its base is `stack/runner-images-base` (= `main` + #53 + #54 + #56) so the diff shows only this PR's own work. Retarget to `main` once those three land.

## What was extracted

| PR | Purpose | LOC (`git diff --shortstat origin/main...<branch>`) |
|---|---|---|
| #53 | Lint the workflows, fix all 23 findings | +105 / −30 |
| #54 | Pin every runner image input to a verified artifact | +202 / −12 |
| #56 | Let Renovate maintain dependencies and action digests | +29 |
| **#43** | Automate pinned runner image publication | **+2,951 / −225** |

#53, #54 and #56 are each based directly on `main` and can merge in any order.

## Honest accounting

`#43` went from **+3,167 → +2,951** insertions and 18 → 10 files. That is only **−6.8%**.

**The whole stack is bigger than the original PR, not smaller**: `git diff --shortstat origin/main...origin/agent/helix-runner-refresh` gives **18 files, +3,249 / −229**, against the original's 18 files, +3,167 / −227. It grew by 82 insertions. Those 82 lines bought two build-breaking bug fixes, two drift tests, a lint gate that found 23 real findings on `main` alone, a fix for CI silently skipping stacked PRs, and Renovate config.

Splitting alone cannot make this PR small, because 2,777 of its 2,951 lines are five files that are mutually dead without each other: the tool (817), its tests (639), the lockfile (557), the workflow (519), and the docs (245). A lockfile nothing reads, a tool nothing calls, and a workflow with no tool are each individually inert, so splitting them further would produce PRs that are smaller but not independently valuable or revertable. **The remaining size is a real cost of the feature, not packaging.** See the review notes at the end for what could still go.

## Why the pins still have ARG defaults

The Dockerfiles keep an `ARG` default for every pin, so `docker build -f docker/codex.Dockerfile .` works off a plain checkout. That is a second copy of every pin.

The hazard was never "the defaults exist" — it was **"a missing pin can silently fall back to a stale default and produce an unpinned image that looks fine."** Deleting the defaults solves that by removing the fallback; asserting them by test solves it by removing the *divergence*, which is the actual hazard. The test version is strictly better on two counts:

1. drift becomes a CI failure with a named cause, instead of a silent stale build;
2. `docker build` keeps working for a human debugging an image by hand.

Deletion was tried and reverted: it would have traded a silent-staleness bug for a "you cannot build this image without first running a tool that refuses to emit args for a local base" bug — worse, because it breaks the debugging path people reach for when something is already wrong. It also *cost* ~30 lines rather than saving any, since `ARG X=value` → `ARG X` is a modification and the fail-closed guards add lines.

Two tests now carry that guarantee, so they were built to hold it:

- they enumerate pins **from the recipe**, never from a hand-written list, so an `ARG` added later is covered the moment it is declared and the guarantee cannot narrow silently;
- failures name the drifting `ARG` and print both values;
- **verified by mutation**: all **42** pinned `ARG` defaults across the six recipes were perturbed one at a time — every one was caught — as was deleting a default outright (12/12).

```
opencode.Dockerfile: ARG CLI_ARM64_SHA512 default has drifted from the resolved pin
  recipe default: 999bf62008bf4e98ad60952ae384c2d028b9cd50a2f50b29c7fecfda90cd729e...
  resolved pin:   727bf62008bf4e98ad60952ae384c2d028b9cd50a2f50b29c7fecfda90cd729e...
```

## Preserved deliberately

- the **"no tag before both smokes"** ordering — push by digest → verify platforms → smoke each arch → attest → tag — unchanged;
- per-backend `fail-fast: false` isolation;
- fail-closed registry absence (only a definitive "no such tag" counts as absent);
- Gemini **not** lockfiled (655-package tree → 8,459-line lock) and Cursor's shell-installer path;
- the Codex `gpt-5.6-luna` / `xhigh` assertion and the opencode `1.18.5` pin.

## One real cut

`build_arguments` had a branch emitting empty strings for absent artifacts. `_lockfile_source` either returns a one-element list or raises, so it could never run. Removed (817 lines, was 824).

## Verification

- `uv run pytest tests/unit -q` → **912 passed**
- `uv run mypy --strict src/helix/ tools/` → clean (30 files)
- `actionlint` → clean, run with **shellcheck 0.11.0 on PATH** (without it, every SC rule is silently skipped)
- `zizmor 1.16.3 --persona=regular` → clean
- `renovate-config-validator` → valid

**No image was built or smoke-tested. There is no Docker daemon on the workstation this was prepared on.** Every claim about image *contents* rests on reading the recipes, not on running them.

## Notes for review

- `build_arguments` rejects any base image that is not `ghcr.io/…@sha256:…`, so it refuses to emit args for `helix-runner-base:proof`. That is correct for CI and hostile to a human; a documented local-build escape hatch is worth considering.
- Neither of the two Dockerfile line-continuation bugs fixed in #54 could ever have built. Nothing in CI builds images on a pull request, so a Dockerfile linter (hadolint) would be the highest-value follow-up gate.
- `docker/runner-versions.json` (59) still overlaps the lockfile and the recipes; collapsing the residual catalog is the next real reduction available.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,10 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - run: uv sync --all-extras
       - run: uv run pytest tests/unit/ -q
-      - run: uv run mypy --strict src/helix/
+      # tools/ is in scope deliberately: until it was, nothing type-checked the
+      # runner-image tool, which is how twenty lines of duplicated dead code in
+      # its main() reached this branch unnoticed.
+      - run: uv run mypy --strict src/helix/ tools/
 
   # Lint the workflows themselves. actionlint checks workflow structure,
   # expression types, and the shell inside every `run:`; zizmor checks the
--- a/.github/workflows/publish-runners.yml
+++ b/.github/workflows/publish-runners.yml
@@ -1,307 +1,602 @@
-name: Publish Runner Images
+name: Refresh Runner Images
+
+# Nightly is the CHECK cadence, not the release cadence.  The workflow asks
+# every upstream what its latest release is; a backend is rebuilt only when
+# that version is not in the registry yet.  A quiet night is a clean no-op.
+#
+# Recipe changes (editing a Dockerfile, bumping the Debian snapshot or a uv
+# wheel) are not upstream releases, so they ship with a manual `force: true`
+# dispatch rather than a push trigger.
 
 on:
-  push:
-    tags:
-      - "v*"
+  schedule:
+    - cron: "17 07 * * *"
   workflow_dispatch:
+    inputs:
+      operation:
+        description: Refresh upstream CLIs or roll one convenience tag back
+        required: true
+        default: refresh
+        type: choice
+        options:
+          - refresh
+          - rollback
+      backend:
+        description: Backend to force-rebuild or roll back
+        required: false
+        type: choice
+        options:
+          - claude
+          - codex
+          - cursor
+          - gemini
+          - opencode
+      digest:
+        description: Existing multi-arch sha256 digest to roll back to
+        required: false
+        type: string
+      force:
+        description: Rebuild even when the upstream version is unchanged
+        required: false
+        default: false
+        type: boolean
+
+concurrency:
+  # Refresh and rollback are serialized independently so an incident response
+  # is never queued behind a long in-flight nightly build.
+  group: runner-image-${{ github.event.inputs.operation || 'refresh' }}-${{ github.repository }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
 
 env:
   REGISTRY: ghcr.io
   IMAGE_PREFIX: ghcr.io/ke7/helix-evo-runner
+  CATALOG: docker/runner-versions.json
+  LOCKFILE: docker/package-lock.json
+  RETENTION_DAYS: 90
+  # gh expects [host/]owner/repo/path, not an https:// URL.
+  ATTESTATION_SIGNER_WORKFLOW: KE7/helix/.github/workflows/publish-runners.yml
+  ATTESTATION_SOURCE_REF: refs/heads/main
 
 jobs:
-  # ── 1. Build base image, one job per arch, push by digest ─────────────────
-  build-base:
-    name: Build base image (${{ matrix.platform }})
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - platform: linux/amd64
-            runner: ubuntu-latest
-          - platform: linux/arm64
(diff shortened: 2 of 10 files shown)


KE7:
Closing: the size is real, not packaging, and the value has been extracted.

A shrink was attempted in earnest and moved the diff from 3,167 to 3,160 insertions — seven lines. Every other cut lost a capability and was reverted.

**The floor is one backend.** Per `discover()`'s own docstring: *"Everything except Cursor is already pinned in git … so this reaches the network only for Cursor, whose version lives inside a shell script at https://cursor.com/install"*. Cursor has no package ecosystem, no Renovate datasource, and its version embedded in an installer, so it can only be tracked by fetching and parsing on a schedule. That generates `resolve_cursor_checksums` / `discover` / `change_plan` (~227 lines), their tests, and the nightly job whose only purpose is to run them — roughly 700 lines to keep one of five backends updating unattended.

**The supply-chain value is already merged-ready elsewhere:** #54 carries digest-pinned bases, exact per-archive Debian Snapshot timestamps, 17 independently verified artifact checksums, and the fix for two Dockerfiles that could never build (`cursor` and `codex`, both now confirmed building with their CLIs executing: `2026.07.20-8cc9c0b` and `codex-cli 0.145.0`). #53 adds the workflow linters, #56 the Renovate config.

**For comparison**, publish workflows in projects at 14k–200k stars: traefik `release.yaml` 139 lines (tag-triggered) plus a separate 29-line scheduled `sync-docker-images.yaml`; gitea `pull-docker-dryrun.yml` 47 lines; coder 113; paperless-ngx 259 (the largest found). None hand-rolls a pin catalog and resolver — immich uses `renovate.json`, paperless `dependabot.yml`.

**Follow-ups worth taking from this work**, each small and independent:
- a ~30-line scheduled refresh workflow, in traefik's shape
- a ~47-line PR dry-run that builds the images without publishing — this catches the unbuildable-recipe class that shipped here undetected, introduced by a commit titled "drop the labels that restate provenance"
- manual Cursor pinning, bumped on notice

Branch `agent/helix-runner-refresh` retained; backups at `backup/pr43-prerebase`, `backup/pr43-preminimize`, `backup/runner-refresh-pre-split-10f8c5f`.

PR #45, opened 2026-08-06
fix(mutator): clean up failed mutations and isolate sandboxed opencode

KE7:
## Scope

Hardens OpenCode mutation execution and ensures a failed mutation cleans up its candidate worktree and branch. It also reports Docker exit 125 as a Docker-invocation failure rather than attempting to parse it as evaluator output.

The credential fix and the `XDG_DATA_HOME` redirect ship together deliberately. The redirect prevents parallel workers contending on one `opencode.db`, but it also moves OpenCode's credential files — so shipping the redirect alone would leave the sandboxed path unauthenticated.

## Fixes

1. **Permission-flag compatibility.** Recent OpenCode releases advertise `--auto`; older releases advertise `--dangerously-skip-permissions`. HELIX probes `opencode run --help`, selects the exact advertised option, and caches it once per process. The cache is lock-protected, so a P×N batch makes one probe even when its first workers start together.

2. **Credential isolation for direct subprocesses.** Per-candidate `XDG_DATA_HOME` prevents P>1 workers sharing `opencode.db`, but it also hides `auth.json` and `account.json`. The direct path symlinks those two files into the candidate state directory: credentials remain reachable, no credential bytes are copied into a worktree, and the database remains per-candidate. The state directory is ignored by Git.

3. **Failure classification.** Structured backend fields are consulted before any text heuristic. `error_max_turns` is handled ahead of the rate-limit check, so a proposal that exhausted its turns is gated as a partial success rather than discarded. Rate-limit detection requires an HTTP status context: a session identifier containing the digits `529` no longer reads as an overload response, and a configuration error mentioning a quota field is not retryable.

4. **Output compatibility.** JSONL parsing is preserved and an older pretty-printed single JSON object is additionally accepted. Malformed structured output still fails loudly.

5. **Failed-mutation cleanup.** Remove the candidate worktree and branch when setup or backend invocation fails, including on unexpected exceptions.

6. **Docker pre-start failures.** Treat Docker exit 125 as an invocation failure and retain Docker's own diagnostic, instead of surfacing it as a missing-`HELIX_RESULT` parse error several frames from the cause.

## Tests

- both permission-flag spellings, exact-option matching, fallback, and cross-thread cache behaviour
- JSONL and pretty-printed single-object structured output
- credentials reachable under isolated XDG state as symlinks, with Git confirming the link is ignored and no credential bytes are copied
- 429 routes to `RateLimitError`; a `529` inside a session identifier and quota-configuration errors do not
- `error_max_turns` preserves the candidate for gating
- sandboxed XDG state remains candidate-scoped

## Validation

- `uv run pytest tests/unit -q` — **909 passed**
- `uv run mypy --strict src/helix/` — **clean**

## Limitation

Container-side credential linking remains unimplemented and unverified. Inside the container, credentials are reachable only through the shared auth volume, so the equivalent link must be made and exercised there.

### DIFF
--- a/src/helix/exceptions.py
+++ b/src/helix/exceptions.py
@@ -114,10 +114,10 @@ class RateLimitError(HelixError):
     logging the failure and returning ``None`` for that proposal slot rather
     than crashing the entire run.
 
-    Detected heuristically from non-zero exit codes or JSON error fields that
-    contain keywords such as "rate limit", "overloaded", "529", "usage limit",
-    or "extra usage".  After retries are exhausted the error bubbles up to the
-    evolution loop, which logs it and continues with a smaller proposal set.
+    Classified first from structured backend status fields, with a narrowly
+    scoped text fallback only when no structured signal is available. After
+    retries are exhausted the error bubbles up to the evolution loop, which
+    logs it and continues with a smaller proposal set.
     """
 
 
--- a/src/helix/executor.py
+++ b/src/helix/executor.py
@@ -345,6 +345,23 @@ def run_evaluator(
             error_ctx,
         )
 
+    # Docker reserves exit 125 for failures before the container command
+    # starts (daemon, image, or invocation errors).  A structured-result
+    # parser cannot add useful information here and would otherwise hide the
+    # actual Docker diagnostic behind "no HELIX_RESULT= line".
+    if returncode == 125:
+        raise EvaluatorError(
+            "Evaluator Docker invocation failed before the evaluator started.",
+            operation="run_evaluator",
+            phase="docker invocation",
+            command=evaluator.command,
+            cwd=str(candidate.worktree_path),
+            stdout=stdout,
+            stderr=stderr,
+            exit_code=returncode,
+            suggestion="Check the Docker diagnostic in stderr above.",
+        )
+
     # Collect ASI
     asi = _collect_asi(
         stdout,
--- a/src/helix/mutator.py
+++ b/src/helix/mutator.py
@@ -719,6 +719,42 @@ def _write_mutation_prompt_artifact(worktree_path: str, prompt: str) -> str:
 # ---------------------------------------------------------------------------
 
 
+# OpenCode credential files, which sit beside the session database inside the
+# opencode data directory.  These must survive per-candidate XDG_DATA_HOME
+# isolation; the database deliberately must not.
+OPENCODE_CREDENTIAL_FILES = ("auth.json", "account.json")
+
+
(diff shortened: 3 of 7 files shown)


KE7:
Closing on the unresolved P1, but the majority of this branch is worth recovering and should not be lost with it.

## Why it closes

`_link_opencode_credentials()` symlinks the user's real `auth.json` and `account.json` into the candidate's state directory. The docstring says *"Symlinks rather than copies, so no credential bytes are written into the candidate worktree"* — but a symlink **resolves**. The OpenCode agent runs with `--auto` and unrestricted workspace access, so it can read those bytes and write them into a tracked file, and `snapshot_candidate()` uses `git add -A`. **`.gitignore` is a default exclusion, not a security boundary.**

The branch also mixes six concerns across 1,081 lines: failed-mutation cleanup, sandbox state redirection, CLI compatibility, output parsing, credential handling, and failure classification.

## What should be recovered — these are real bugs found by running the thing

| commit | keep? | what it fixes |
|---|---|---|
| `1f01480` | **yes** | `--auto` flag. OpenCode 1.18.5 rejects `--dangerously-skip-permissions`; the CLI exits before doing any work. Version-probed and cached, so both spellings work. |
| `665ba19` | **yes** | `--format json` emits JSONL, not one object. `json.loads(stdout)` fails outright. Accepts both shapes. |
| `f6af1e7` | **yes** | the flag-probe cache was not concurrency-safe under P×N. |
| `952f53b` | **yes** | failure misclassification, four instances. The worst: `_looks_like_rate_limit` matched `"529"` anywhere in stdout, including inside a session UUID, and the rate-limit guard runs **before** the `error_max_turns` handler that treats turn exhaustion as partial success. A proposal that had completed 60 turns of real work was discarded on a hex collision — roughly 0.7% of mutations, scaling with P×N. |
| `d16520b` | **partial** | failed-mutation cleanup is good; the XDG widening from unsandboxed-only to all runs is what makes the credential problem reach the sandboxed path. |
| `4f53fbb` | **no** | the credential symlink. This is the blocker. |

Note the dependency: `d16520b` widens the `XDG_DATA_HOME` redirect (correctly — it prevents concurrent workers colliding on one SQLite WAL), and that redirect is *why* credentials become unreachable. Recovering `d16520b` without `4f53fbb` leaves OpenCode unauthenticated. So a follow-up PR should either take the four compat/classification commits alone, or pair the redirect with a credential mechanism that does not place a resolvable path to host credentials inside an agent-writable workspace.

Branch `fix/mutation-worktree-cleanup` retained; nothing is lost.

PR #52, opened 2026-08-10
fix(circle_packing): stop the evaluator inheriting an arbitrary python3

KE7:
The circle_packing example scored **0.0 for every candidate — including its own reference solution** — so the documented 2.6360 result was not reproducible.

## Root cause

Not the solver. The evaluator command was `python3 evaluate.py`, which resolves to whatever is first on `PATH`. Here that is the Xcode 3.9 interpreter, with a user-site numpy that trips numpy's own *"do not import from the source directory"* guard and no scipy at all.

`evaluate.py` catches the `ImportError` and reports `score: 0.0` — **indistinguishable from a genuinely bad candidate**. An entire evolution run would look like it simply made no progress.

`solve_optimized.py` additionally uses PEP 604 `X | None` annotations, so even a 3.9 interpreter with a working numpy fails with a `TypeError`.

## Fix

```
uv run --no-project --python 3.12 --with numpy --with scipy python3 evaluate.py
```

`--no-project` is deliberate: a fresh clone reproduces the benchmark without syncing the repo environment, and HELIX does not acquire a scipy dependency for one example's sake.

## Verified end to end

| | before | after |
|---|---|---|
| `solve_optimized.py` (reference) | 0.0 | **2.635982**, 0 violations |
| `solve.py` (naive seed) | 0.0 | **0.979764**, 0 violations |

circle_packing was the only example still using a bare `python3`; `web_researcher` already used `uv run`.

### DIFF
--- a/examples/circle_packing/README.md
+++ b/examples/circle_packing/README.md
@@ -51,3 +51,22 @@ No Sonnet, no Opus, no extended thinking — just Haiku with low effort and a ha
 - `evaluate.py` — scorer that checks validity and returns `Σ rᵢ`.
 - `helix.toml` — project configuration.
 - `solve_optimized.py` — a hand-tuned reference implementation that scores **2.635982**. It is not used during evolution; it is provided as a sanity check / target for comparison.
+
+## Requirements
+
+The solver uses `numpy` and `scipy` and requires Python 3.10+ for PEP 604
+(`X | None`) annotations. The evaluator command in `helix.toml` pins an
+interpreter and installs those two packages itself via `uv`, so no manual
+setup is needed and a fresh clone reproduces the reference score.
+
+Verify the baseline before trusting a run:
+
+```bash
+cd $(mktemp -d) && cp <repo>/examples/circle_packing/evaluate.py . \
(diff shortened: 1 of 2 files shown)


KE7:
Superseded by #49. The complete #52 change was recovered onto #49's branch in commit [acd5681](https://github.com/KE7/helix/commit/acd5681).

| File | Recovery |
|---|---|
| `examples/circle_packing/README.md` | Carried all 19 added lines: the Python 3.10+/PEP 604 and numpy/scipy requirements, the pinned `uv run --no-project --python 3.12 --with numpy --with scipy python3 evaluate.py` verification command, the fresh-clone reproducibility explanation, and the diagnostic distinguishing an ImportError-induced 0.0 from a bad candidate. |
| `examples/circle_packing/helix.toml` | Carried the full explanatory comment above `command` and replaced bare `python3 evaluate.py` with `uv run --no-project --python 3.12 --with numpy --with scipy python3 evaluate.py`. #49's adjacent `score_parser = "helix_result"` and dataset sizing were retained. |

The reference evaluator now emits a real score (2.635982, 0 violations), so this interpreter fix is verified on #49. The #52 branch is retained.

PR #54, opened 2026-08-10
chore: add Renovate configuration for automated dependency updates

KE7:
## What

Adds a 30-line Renovate configuration: weekly Monday dependency updates, at most 5 concurrent PRs, semantic commits, GitHub Action digest pinning, digest pinning for the runner base image, and immediate vulnerability alerts.

This PR is Renovate-only. The weekly `publish-runners` rebuild that used to ride along here now lives in PR #71 (together with the node-user image verification and the `resolve-tag` job), so `.github/workflows/publish-runners.yml` is untouched by this diff.

## Why

Keep dependency maintenance predictable while retaining immutable references for Actions and the base image.

## Review focus

- Renovate: the `dockerfile` digest-pinning rule is scoped with `matchFileNames` to `docker/base.Dockerfile` (the only Dockerfile that `FROM`s a registry image). The other five Dockerfiles start with `ARG BASE_IMAGE=helix-runner-base:latest` / `FROM ${BASE_IMAGE}`, which Renovate would resolve to a nonexistent Docker Hub library image and emit lookup warnings for; a second rule disables the `dockerfile` manager for `docker/*.Dockerfile` minus `base.Dockerfile`. Their agent CLI installs (`npm install -g` latest, `curl | bash` for cursor) stay unpinned by design — the weekly rebuild in #71 is what tracks those, and the dockerfile manager would never have touched them anyway (it only extracts `FROM`/`COPY --from` images).
- Neither rule overrides Renovate's recommended defaults beyond what the rule needs.

## Validation

- `renovate.json` parses as JSON; 30 lines, 2 package rules, and a +30/−0 configuration-only diff.
- `renovate-config-validator --strict renovate.json` (renovate 44.74.1): `Config validated successfully`.
- `git diff origin/main --stat` after merging `main`: only `renovate.json`.

### DIFF
--- a/.github/workflows/publish-runners.yml
+++ b/.github/workflows/publish-runners.yml
@@ -4,6 +4,14 @@ on:
   push:
     tags:
       - "v*"
+  schedule:
+    # Weekly rebuild so the images pick up whatever the unpinned agent CLI
+    # installs (npm latest / curl-installer latest) resolve to, without
+    # waiting on a release tag. 12:00 UTC Monday is before 6am in
+    # America/Los_Angeles year-round (04:00 PST / 05:00 PDT), matching the
+    # cadence and timezone this repo's Renovate config already uses
+    # ("before 6am on monday", "timezone": "America/Los_Angeles").
+    - cron: "0 12 * * 1"
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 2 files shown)


KE7:
Closing: the weekly runner-image rebuild now lives in #71, and the remaining Renovate config (base-image digest and Actions digest pinning) is not needed right now.

PR #55, opened 2026-08-10
fix(config): make minibatch_size default mode-aware

KE7:
## Summary

`evolution.minibatch_size` now defaults to **1** in single-task / no-example
mode and **3** in multi-task mode, while preserving every explicit value. The
top-level `model_validator(mode="before")` rewrites the raw mapping before
`EvolutionConfig` resolves `num_parallel_proposals = "auto"`, preventing a
stale derived proposal width.

Boolean dataset cardinalities are rejected, so Pydantic cannot coerce `true`
to `1` and select the wrong mode.

`"auto"` is supported correctly whether or not a checked-in example uses it.
The end-to-end unit test drives the real `run_evolution` proposal loop,
asserting both the resolved `P` and the number of planned proposal slots.

## Why the default is applied to the raw mapping

Pydantic's `model_copy(update=...)` deliberately skips validation, so patching
a built config does not re-resolve anything downstream of the patched value —
and by that point both the implicit `minibatch_size` default and the resolved
`"auto"` proposal width are concrete. Switching dataset modes therefore
requires rebuilding from raw config.

## Behaviour

The user-visible effect remains limited to single-task configurations using
`num_parallel_proposals = "auto"`, which resolves as
`max(1, max_workers // minibatch_size)`. For example, `max_workers=12` changes
from 4 to 12 slots, `8` from 2 to 8, and `2` from 1 to 2. Proposal-width growth
depends on integer flooring and the minimum-one clamp, so it is not always
threefold. Integer proposal widths are unchanged, and single-task metric-call
counts are unchanged because no train sampler is constructed.

## Verification

- `uv sync --extra dev`
- `uv run pytest tests/unit -q` — 885 passed
- `uv run mypy --strict src/helix/` — clean
- `uv run ruff check src/helix/config.py tests/unit/test_minibatch_mode_default.py`
- `uv run ruff format --check src/helix/config.py tests/unit/test_minibatch_mode_default.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- **BEHAVIOUR CHANGE — `evolution.minibatch_size` default is now mode-aware.**
+  When `minibatch_size` is omitted it resolves to **1** in single-task /
+  no-example mode (no `[dataset].train_size` and no `[seedless].train_path`)
+  and stays **3** in multi-task / generalization mode. Previously every mode
+  got a flat 3, which was only ever the multi-task half of the intended
+  default. Single-task is HELIX's default mode, so this affects existing
+  configs that relied on the implicit 3 — including `examples/circle_packing`.
+
+  **The change is user-visible only through `num_parallel_proposals = "auto"`,
+  and there it triples proposal width.** `"auto"` derives
+  `max(1, max_workers // minibatch_size)`, and the resolved `P` drives the
+  proposal-slot loop. A single-task run with `max_workers = 12` and
+  `"auto"` now opens **12** proposal slots per generation where it previously
+  opened 4 — verified end-to-end against the evolution loop. More proposals
+  per generation means more mutation agent invocations per generation, so
+  review `max_workers` before your next single-task `"auto"` run.
+
+  Single-task runs that pin `num_parallel_proposals` to an integer see no
+  behavioural change: with no training split HELIX builds no batch sampler, so
+  `minibatch_size` is never read for sampling and metric-call counts are
+  unaffected. For those configs this is a parity/consistency fix only.
+
+  An explicitly written `minibatch_size` is never rewritten in either
+  direction: set `minibatch_size = 3` to keep the previous behaviour exactly.
+
 ## [0.2.2] - 2026-05-13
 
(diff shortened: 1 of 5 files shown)


KE7:
Closing: correct, requested, and currently unexercised.

The mode-aware default is right — upstream practice is 1 for single-task and 3 for multi-task, and HELIX hard-coded 3. The implementation is also careful: a `model_validator(mode="before")` on `HelixConfig` rewrites the raw mapping so `EvolutionConfig` is constructed correctly and `"auto"` resolves from the corrected value, avoiding the ordering trap where a later mutation leaves a stale derived P. That was verified by building the naive version and watching 5 of 17 tests fail against it.

The problem is that **its only observable effect has no consumer today**. In single-task mode `minibatch_size` is never read for sampling — `batch_sampler` is constructed only when a train loader exists — so the sole behavioural consequence is auto-width resolution, `max(1, max_workers // minibatch_size)`. **No checked-in example sets `num_parallel_proposals = "auto"`**, so nothing currently exercises the path this corrects. That is 378 lines for a latent fix.

Four review findings also remain open at head: `dataset.train_size = true` selects the wrong mode via bool/int coercion; `model_copy(update=...)` can retain a stale implicit default; the CHANGELOG says width always "triples" which is false for non-multiples and floored widths; and the new test file has a Ruff F811.

**Re-open when a config actually uses `"auto"`.** The durable core is roughly the 40-line raw-input validator plus targeted regressions; the surrounding documentation and test expansion can be left behind.

Branch `fix/minibatch-mode-aware-default` retained.

KE7:
Reopening. Karim's call: rather than defer until something uses `"auto"`, make a config use it — which turns the latent fix into an exercised one and gives the PR its missing consumer.

Two things to land before this merges:
1. A checked-in example config that sets `num_parallel_proposals = "auto"`, so the auto-width path this PR corrects is actually driven end-to-end rather than only unit-tested at the config layer.
2. The four open review findings: `dataset.train_size = true` selects the wrong mode through bool/int coercion; `model_copy(update=...)` can retain a stale implicit default; the CHANGELOG's claim that width always "triples" is false for non-multiples and floored widths; and the new test file has a Ruff F811.

KE7:
Closing this PR: #48 removed `num_parallel_proposals = "auto"`, which was the only user-visible consequence of the mode-aware default. The focused tests show the mismatch directly: 7 fail because `"auto"` is no longer accepted and 11 pass. This change should not merge.

PR #57, opened 2026-08-10
fix(auth): never auto-forward an Anthropic API key into agent containers

KE7:
## Summary

This PR completes the login-safe authentication path for Anthropic-backed agents.

- Stops automatic forwarding of `ANTHROPIC_API_KEY` and `ANTHROPIC_AUTH_TOKEN`; login is the default and an API key requires an explicit `[env]` or `passthrough_env` opt-in.
- Preserves non-secret `USER` and `LOGNAME` for every agent CLI by default. On macOS, Claude Code needs `USER` to resolve the stored Keychain login; evaluator scrubbing remains unchanged.
- On an authentication failure, reports the names (never values) of environment variables that actually reached the agent.

## Root cause and user impact

The original scrub reduced an agent environment to `PATH`, `HOME`, and `HELIX_*`. Claude Code's macOS Keychain lookup requires `USER`, so an otherwise valid stored login appeared as:

```
Not logged in · Please run /login
```

Combined with the key-forwarding defect this PR fixes, that produced a dangerous split:

| User state | Previous result |
| --- | --- |
| Logged in, no API key | “Not logged in” |
| Logged in, ambient API key | Run succeeds using the API key, silently bypassing the login |

Neither case made the paying credential clear. In practice an API key was effectively mandatory for the Claude backend—sandboxed and unsandboxed—even though the documented login path existed. This PR makes login functional and prevents an ambient key from silently taking over.

`USER` is now a default rather than a documented opt-in because requiring users to discover an OS-specific identity variable would recreate the same authentication trap. `LOGNAME` is included as the conventional companion used by Unix-like tools; both are non-secret. Applying this narrowly to all agent CLIs also avoids the same hidden login-resolution dependency in Codex, Cursor, Gemini, or OpenCode, while keeping evaluator environments strictly scrubbed.

## Explicit API-key opt-in

An API key remains available only when deliberately configured:

```toml
# Either forward an explicit value:
[env]
ANTHROPIC_API_KEY = "sk-ant-..."

# Or pass the ambient one through. Note this is TOP-LEVEL: placing it under
# [env] would parse as env.passthrough_env and silently do nothing.
passthrough_env = ["ANTHROPIC_API_KEY"]
```

When chosen, HELIX warns that the key takes precedence over the stored login.

## Validation

- New tests were run first and failed before the fix: missing `USER` in the real Claude subprocess environment, preserved explicit passthrough, and absent auth-failure environment diagnostic.
- `uv run pytest tests/unit -q` → 884 passed
- `uv run mypy --strict src/helix/` → Success: no issues found in 29 source files
- `uv run ruff check src/helix/backends.py src/helix/mutator.py tests/unit/test_backend_auth_env.py` → clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,47 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING (behaviour):** `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` present in
+  your shell are no longer forwarded into agent subprocesses or agent containers.
+  Anthropic backends (`claude`, `opencode`) authenticate through
+  `helix sandbox login <backend>`; an API key in the agent environment silently
+  overrode that login, so a user who had just logged in ran on the key instead
+  without being told. HELIX now resolves the ambiguity in favour of login.
+
+  Before — key auto-forwarded, `helix sandbox login claude` silently ineffective:
+
+  ```console
+  $ export ANTHROPIC_API_KEY=sk-ant-...
+  $ helix sandbox login claude   # credentials written to the helix-auth-claude volume
+  $ helix evolve                 # agent runs on the API key; the login is ignored
+  ```
+
+  After — the login is used, and the dropped key is announced:
+
+  ```console
+  $ helix evolve
+  WARNING  ANTHROPIC_API_KEY found in the environment but NOT forwarded to Claude
+           Code: Anthropic backends authenticate via `helix sandbox login claude` ...
+  ```
+
+  **Migration.** Either run `helix sandbox login <backend>` (recommended), or opt
+  the key in explicitly in `helix.toml`:
+
+  ```toml
+  [env]
+  ANTHROPIC_API_KEY = "sk-ant-..."
+  # or, to take the value from your shell:
+  # passthrough_env = ["ANTHROPIC_API_KEY"]
+  ```
+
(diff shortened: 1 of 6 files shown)


KE7:
Superseded by #61, which now carries everything unique to this PR.

The two PRs answered the same question — how credentials reach an agent — and the split was an accident of timing: this one was written before the two-mode design existed.

**Subsumed:** the `BACKEND_AUTH_ENV` pruning here removed the `claude` entry and `ANTHROPIC_API_KEY` from `opencode`. #61 deletes the table entirely, which is the point of the two-mode design, so the careful pruning is moot.

**Folded into #61 (commit `e803d5c`), because #61 lacked both:**
- `AGENT_LOGIN_IDENTITY_ENV = ("USER", "LOGNAME")`. Load-bearing, not cosmetic: HELIX's environment scrub reduces the agent environment to `PATH`, `HOME` and `HELIX_*`, and on macOS the Claude CLI resolves its stored login through the Keychain, which requires `USER`. Without it every mutation fails in ~50 ms with "Not logged in" while the identical command works by hand — found by bisection during a live run after it had already killed an arm. #61 implements login-based auth, so shipping it without this would have meant shipping a login path that cannot resolve a login.
- `_looks_like_authentication_failure` and the diagnostic reporting the **names** (never values) of variables that actually reached the agent.

Verified on #61 at `e803d5c`: both present, `BACKEND_AUTH_ENV` still absent, evaluator subprocesses still receive their original strict environment, 880 unit tests and strict mypy green.

Branch `fix/no-anthropic-key-forwarding` retained.

PR #61, opened 2026-08-11
Isolate sandbox login credentials per candidate; replace gemini backend with agy

KE7:
## Summary

This PR does two things together, because the second was built directly on the first's per-candidate auth design and reuses its manifest:

- **Isolates sandbox login credentials per candidate.** `sandbox.auth = "volume"` now seeds a private, per-candidate Docker volume from an allowlist of credential files in the operator's `helix-auth-<backend>` login volume; no agent container ever mounts that shared volume. Both concurrent and sequential candidates get disjoint writable auth state, and the candidate volume is removed after the candidate finishes.
- **Removes the `gemini` mutation backend and replaces it with `agy`** (Google's Antigravity CLI). **This is breaking**: `agent.backend = "gemini"` is now rejected. `gemini`'s runner image, Dockerfile, sandbox auth manifest entry, and CI matrix rows are removed outright, not deprecated; see the `CHANGELOG.md` entry for the migration note. **This must land together with the pending `0.2.1` → `0.3.0` version bump.**

Also folded in: seeding Cursor from the credential file its CLI actually reads rather than its settings file, sidecar-only host values through `evaluator.sidecar.passthrough_env` kept disjoint from agent credentials, confining `USER`/`LOGNAME` to the unsandboxed path, naming the cause when a credential seed refuses (including the everyday case — nobody has signed in yet), and having CI's `docker-integration` job build its own fixture images from this checkout's Dockerfiles instead of pulling published `:latest` tags.

## Real-grant verification

Four of the five backends are now confirmed end to end against a real grant, not a synthetic record. Each was seeded from its real, already-authenticated login volume through HELIX's own candidate-volume path (`run_sandboxed_command(..., scope="agent", auth="volume")` → volume creation → seeding → the agent's own Docker argv → cleanup), then checked from inside the candidate container with the backend's own free, no-spend status command:

| Backend | Status command | Result |
| --- | --- | --- |
| Claude | `claude auth status --text` | exit 0, `Login method: Claude Max account` |
| Codex | `codex login status` | exit 0, `Logged in using ChatGPT` |
| Cursor | `cursor-agent status` | exit 0, signed-in confirmation naming the account |
| OpenCode | `auth list` | exit 0, `1 credentials` |

These checks ran under Docker's amd64-on-arm64 emulation (no arm64 runner image exists yet); all four still passed, and HELIX's CI already runs these images as amd64, so the confirmed path is the one that ships.

**`agy`'s manifest entry is derived, not verified.** Its credential path was traced with `strace` against a real, unauthenticated Linux binary, but its login volume holds no credential yet, so nothing has been seeded or probed, and the credential file's JSON shape is still unknown. It ships enabled anyway, the same as every other manifest entry — see `docs/design/sandbox-auth-projection.md` for why that's a documentation standard, not a code gate, and for the mechanism a missing/invalid credential actually hits (a closed-by-default seed failure naming the cause and pointing at `helix sandbox login <backend>`).

Also folds in and supersedes the now-closed backend-swap PR (#66), which is why this diff includes the `agy` migration rather than only credential isolation.

## Verification
- `uv sync --extra dev`
- `uv run python -m pytest` — 982 passed
- `HELIX_DOCKER_TESTS_STRICT=1 uv run python -m pytest tests/integration/ -m docker_integration -q -rs` — 27 passed, 0 skipped (one fewer than this table's historical 28: `gemini`'s dedicated auth-method-gate test was removed with no direct replacement, since exercising `agy`'s equivalent live would require a `--print` call this work's hard limits forbid)
- `uv run ruff check src/ tests/` — clean
- `uv run mypy --strict src/helix/` — clean
- Diff: +2935/-396 lines across 24 files

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,9 +22,16 @@ jobs:
   docker-integration:
     # Real-Docker proof that per-candidate auth volumes actually isolate
     # credentials. Every credential used is synthetic, so this job needs no
-    # backend secrets: the runner images are anonymously pullable from GHCR,
-    # and the seeding helper runs in the backend's own runner image rather
-    # than introducing a third-party one.
+    # backend secrets.
+    #
+    # The fixture images are built from this checkout's own Dockerfiles
+    # rather than pulled from GHCR: a pulled image only reflects the last
+    # *published* build, so a PR that edits a Dockerfile would be tested
+    # against code it isn't changing, and a PR that adds a new backend could
+    # never go green -- its image can't exist on GHCR until after it merges.
+    # Building locally means every PR is validated against exactly the
+    # images its own diff produces. The seeding helper still runs in the
+    # backend's own runner image rather than introducing a third-party one.
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
@@ -32,13 +39,23 @@ jobs:
         with:
           python-version: "3.12"
       - run: uv sync --all-extras
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
       # The tests skip when a fixture image is absent. A fresh runner has
-      # none, so without this pull the job would pass while running nothing.
-      - name: Pull fixture images
-        run: |
-          for b in claude codex cursor gemini opencode; do
-            docker pull -q "ghcr.io/ke7/helix-evo-runner-$b:latest"
-          done
+      # none, so without this build the job would pass while running
+      # nothing. `docker buildx bake` (see docker/docker-bake.hcl) builds
+      # the base image once, then every backend's Dockerfile against it in
+      # parallel, using Buildx's named-context feature to point each
+      # backend's `FROM` straight at the base build's own output instead of
+      # a registry -- so build order only matters between the base and the
+      # backends, never among the backends themselves. Each target loads
+      # into the local Docker engine (`--load`, no registry push) under the
+      # exact tag `DEFAULT_BACKEND_IMAGES` in `src/helix/backends.py`
+      # resolves, and each caches independently through the GitHub Actions
+      # cache so an unrelated change doesn't invalidate every other image's
+      # layers.
+      - name: Build fixture images
+        run: docker buildx bake -f docker/docker-bake.hcl --load
       # STRICT=1 turns any remaining skip into a failure, so this job can
       # never be green without actually exercising the isolation tests.
       - run: uv run pytest tests/integration/ -m docker_integration -q -rs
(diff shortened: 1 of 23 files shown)


KE7:
Closing: the central design here is wrong, and the parts worth keeping have landed elsewhere.

## Why the per-candidate credential volume was dropped

Giving each candidate a private copy of the operator's credential, destroyed with the candidate, breaks
token refresh in two ways that only appear on long runs:

- **No write-back path exists.** A refresh performed inside a candidate is discarded with that
  candidate's volume, so the login volume never receives the refreshed credential. Later candidates are
  seeded from an ageing one. Once the stored refresh token is spent, the login volume keeps the spent
  one indefinitely — and `helix sandbox status` still reports success for backends where it is a
  file-existence probe.
- **It relocates the CLI's own refresh lock.** Claude Code serialises refresh across processes with a
  lockfile beside the credential, plus a double-checked re-read. Under a private per-candidate volume
  that lock serialises nothing, so with parallel proposals several candidates race one refresh token
  once the seeded access token passes the CLI's five-minute expiry buffer.

The shared login volume this replaces has powered multi-day runs. Measurement also showed the shared
file's own refresh bookkeeping acts as an accidental serialiser — so isolation would have removed a
protection rather than adding one.

## What landed instead

- The `gemini` to `agy` backend migration, and CI building its own fixture images rather than pulling
  published ones, moved to #71.
- The problem this branch was really trying to solve — candidates inheriting each other's transcripts,
  memories, goals and session databases — is addressed in #72 by relocating agent **state** to a
  per-candidate directory outside `$HOME` while the credential stays shared exactly as it was.

## Deliberately not carried over

The tmpfs ancestor-ownership helper and the transcript-bind ordering fix were checked and are **not**
independent: both fix conditions that arise only under this branch's tmpfs-`$HOME` and deeper-mount
topology. On the current default mount they have no consumer.

## Findings worth keeping

- Credential and state are separable by configuration for codex (`sqlite_home`, `log_dir`), opencode
  (`OPENCODE_DB`, not `XDG_DATA_HOME`, which moves both) and cursor (`CURSOR_CONFIG_DIR`).
- Claude's config directory knob is all-or-nothing and setting it pulls additional state into the
  shared directory; its refresh lock is a directory, so sharing the lock alone is not possible.
- Codex has no cross-process refresh lock. Simultaneous candidates sharing one credential file can
  race a single-use refresh token, and the losers fail silently at any log level.
- A backend's session database can itself hold access and refresh tokens, so "share only the
  credential file" does not always mean "share only credentials".

The branch is retained.

PR #65, opened 2026-08-16
fix(mutator): classify structured backend failures before text heuristics

KE7:
## What

Non-zero backend exits are classified from their structured result before text heuristics. `error_max_turns` remains partial success, structured HTTP 429 becomes a rate-limit error, and the raw fallback scans both streams with token-aware rate-limit signals. Evaluator exit 125 surfaces a Docker diagnostic only when Helix invoked Docker.

## Why

Avoid discarding completed work because incidental text resembles a quota signal, while retaining a useful error for real Docker invocation failures.

## Review focus

Check structured-envelope precedence over the raw-text fallback, especially the `error_max_turns` path and both quota word orders.

## Validation

`uv run --extra dev pytest tests/unit/test_backend_failure_classification.py tests/unit/test_executor.py -q` — 24 passed in 0.23s. Diff: +492/−66 lines.

## Review fixes

Addresses the verified findings against real Claude Code envelope shapes (`subtype` is always present; success envelopes carry `result: <str>`, error subtypes carry `errors: [...]`; there is no top-level `error` field; `api_error_status` only appears on newer CLIs' success envelopes).

- **Envelope classification no longer keys on `subtype` presence** (`src/helix/mutator.py`). `_has_structured_failure_signal` is replaced by `_envelope_reports_failure` (`is_error` or an `error_*` subtype). Only then are the backend's own strings — `errors` and `result` — classified for rate-limit/overload tokens before falling back to a generic error. `{"subtype":"success","is_error":true,"result":"API Error: 429 ..."}` and `{"subtype":"error_during_execution","errors":["API Error: 529 overloaded_error"]}` now raise `RateLimitError`; a successful turn whose prose mentions rate limits is returned as a completed mutation.
- **Reads `errors`/`result`, not `error`.** Fixtures on the branch and inherited ones (`tests/unit/test_resume.py`, `tests/unit/test_rate_limit_fixes.py`) now use the real shapes.
- **Regex tokens**: `overloaded` and `rate[ _-]?limit` use letter/digit lookarounds instead of `\b`, so `overloaded_error` and `rate_limit_error` match.
- **Raw fallback narrowed**: stderr-first precedence restored (`stderr or <stdout view>`). JSONL backends are classified from the parsed events' error fields only (`_events_failure_text`), and the raw stdout view is limited to non-JSON lines — the transcript (tool output, file contents) is never scanned. A parsed claude envelope is never re-scanned as raw text. `test_stdout_rate_limit_is_seen_when_stderr_is_non_empty` is rewritten as `test_jsonl_error_event_rate_limit_is_seen_when_stderr_is_non_empty`, with companion tests that transcript text and stderr precedence stay generic.
- **Exit 125 under Docker** (`src/helix/executor.py`): raised as a Docker invocation failure only when no `HELIX_RESULT=` line was emitted or stderr carries a daemon diagnostic; an evaluator's own exit 125 with a valid result line is scored (aggregate zeroed, per-example scores kept).

Branch was merged with `main` first (clean merge, `417f5ee`).

Gates (post-fix): `uv run python -m pytest -q` — 1032 passed; `uv run ruff check src/ tests/` — All checks passed; `uv run mypy --strict src/helix/` — Success: no issues found in 27 source files.

### DIFF
--- a/src/helix/exceptions.py
+++ b/src/helix/exceptions.py
@@ -114,10 +114,10 @@ class RateLimitError(HelixError):
     logging the failure and returning ``None`` for that proposal slot rather
     than crashing the entire run.
 
-    Detected heuristically from non-zero exit codes or JSON error fields that
-    contain keywords such as "rate limit", "overloaded", "529", "usage limit",
-    or "extra usage".  After retries are exhausted the error bubbles up to the
-    evolution loop, which logs it and continues with a smaller proposal set.
+    Classified first from structured backend status fields, with a narrowly
+    scoped text fallback only when no structured signal is available. After
+    retries are exhausted the error bubbles up to the evolution loop, which
+    logs it and continues with a smaller proposal set.
     """
 
 
--- a/src/helix/executor.py
+++ b/src/helix/executor.py
@@ -233,11 +233,10 @@ def run_evaluator(
 
     evaluator = config.evaluator
     sandbox_image = None
-    # The single honest signal that HELIX itself launched Docker for this
-    # evaluation: we chose the sandboxed branch below.  Evaluator commands
-    # are arbitrary user commands, so the exit code alone (or any text in
-    # it) cannot be trusted to mean "Docker was involved" — only the branch
-    # HELIX actually took can.
+    # Whether HELIX itself launched Docker for this evaluation.  Evaluator
+    # commands are arbitrary user commands, so neither the exit code nor the
+    # command text establishes that Docker was involved; only this branch
+    # does.
     docker_invoked = config.sandbox.enabled and config.sandbox.evaluator
     if docker_invoked:
         if evaluator.sidecar is None:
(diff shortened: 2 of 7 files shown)


KE7:
Closing in favour of #72, which covers this ground with less machinery.

Three reasons this branch is not the right vehicle:

1. **The token fix nets to zero.** `main` matches rate-limit keywords as plain substrings, so `"overloaded"` already matches `overloaded_error` and `"529"` already matches a bare status. This branch rewrote that as regexes with word boundaries, which broke those tokens, and then re-added lookarounds to unbreak them.

2. **The exit-125 handling is self-inflicted.** `main` has no such check. This branch introduced one, review found it fires on an evaluator's own exit 125 and discards a valid `HELIX_RESULT`, and the fix added five hardcoded Docker stderr strings to guard it.

3. **It duplicates #72.** Both branches independently built structured-error extraction over the same envelopes, because this one sat unmerged while #72 needed it. They conflict in four hunks in `mutator.py`.

The one finding worth keeping is real: on a non-zero exit `main` scans `result.stderr or result.stdout`, and for the JSONL backends stdout is the agent's entire transcript, so a candidate working on a repo whose source contains "rate limit" (this repo's own `mutator.py` does) can turn any unrelated failure into a rate-limit verdict. That is being carried into #72 on top of its `_structured_error_texts`: read stderr and the structured error fields, never the transcript.

Not carried over: the regex rework, the quota/usage prose patterns, and the exit-125 machinery.

PR #66, opened 2026-08-24
Replace gemini mutation backend with agy (Antigravity CLI)

KE7:
## Summary

Google's `gemini-cli` backend is being retired; this replaces it with Google's Antigravity CLI (`agy`) as a HELIX mutation backend. Backend id is `agy`, matching HELIX's existing convention of backend ids equal to the CLI command name (`claude`, `codex`, `cursor`, `opencode`).

**⚠ Stacks on #61 (`impl/auth-two-modes`) and must merge after it.** This branch is built directly on top of #61's `AUTH_CREDENTIAL_MANIFEST` / `AUTH_MOUNT_DESTINATIONS` / `AUTH_SYNTHESIZED_SIBLINGS` per-candidate auth design; branching from `main` instead would have reinvented a `gemini` manifest entry this PR immediately deletes. Until #61 merges, this PR's diff against `main` includes #61's commits too — that's expected for a stacked branch, not scope creep.

**⚠ BREAKING.** `agent.backend = "gemini"` is rejected. `gemini`'s runner image, Dockerfile, sandbox auth manifest entry, and CI matrix rows are removed outright, not deprecated. See the CHANGELOG entry.

**⚠ Must land together with the pending 0.2.1 → 0.3.0 version bump**, which is out of scope here on purpose (to avoid racing it). This is no longer a CI-sequencing risk: `ci.yml`'s `docker-integration` job now builds every fixture image (including `agy`'s) from this checkout's own Dockerfiles rather than pulling a published tag, so it doesn't depend on `ghcr.io/ke7/helix-evo-runner-agy` existing yet. The version bump still matters for `publish-runners.yml`, which is the thing that actually publishes `agy`'s image for real HELIX users to run against.

## What changed

- **`src/helix/backends.py`** — `BackendName`/`BACKENDS`: `gemini` → `agy`. `agy` added to `EFFORT_AWARE_BACKENDS` and `EFFORT_VALID_VALUES` (`low`/`medium`/`high`, same set as `claude`) — a real behavioral change from `gemini`, which silently dropped `agent.effort`. New `BACKEND_AUTH_COMMANDS["agy"]`: login is the bare interactive launch (`["agy"]`, no dedicated login subcommand, same pattern as `opencode`); status is a credential-file probe (`agy models` exits 0 even when logged out, so it can't signal status); logout removes only `~/.gemini/antigravity-cli`, never the whole `~/.gemini` (shared with legacy `gemini-cli` state).
- **`src/helix/sandbox.py`** — New `AUTH_CREDENTIAL_MANIFEST["agy"]` and `AUTH_MOUNT_DESTINATIONS["agy"]`, derived from a live `strace` against the real, unauthenticated Linux CLI (path: `.gemini/antigravity-cli/antigravity-oauth-token`; mount: `/home/node/.gemini/antigravity-cli`). The `gemini` `AUTH_SYNTHESIZED_SIBLINGS` entry is deleted with `gemini`, and **no `agy` entry is added** — see "Open unknowns" below. `_synthesized_ancestor_tmpfs_args` needed no changes; it's already generic (verified, see report below).
- **`src/helix/mutator.py`** — New `agy` branch in `_build_backend_args` (`--dangerously-skip-permissions --print --output-format json`, plus `--model`/`--effort` when set), dispatched through the same single-JSON-object parsing branch as `claude` (`_parse_json_object_output`), not the JSONL branch the other three backends use. The `gemini`-specific stream-preamble tolerance, its transcript tool-event counter, and its `GEMINI_CLI_TRUST_WORKSPACE` env special-case are all deleted with no `agy` replacement — `agy` has no confirmed equivalent need, and `_normalise_usage_stats`'s generic walk still produces best-effort usage stats without a dedicated counter.
- **`src/helix/cli.py`** — `sandbox_login`'s gemini-specific hint replaced with an `agy` one; `evolve --backend`'s hardcoded `click.Choice` list and the `--effort` help text updated.
- **`docker/agy.Dockerfile`** (new, `docker/gemini.Dockerfile` removed) — modelled on `cursor.Dockerfile`'s `curl | bash` pattern (agy isn't published to npm). Built and run locally against both `helix-runner-base` and as uid 1000 — see report below; a straight symlink into `/root/.local/bin` (root-only, `700`) silently broke under the uid-1000 runtime user, so the binary is copied to `/opt` first, same fix `cursor.Dockerfile` already applies to its own root-owned install.
- **`.github/workflows/publish-runners.yml`** — matrix swap (`gemini` → `agy`). **`ci.yml`** — unchanged by this PR directly; it inherits #61's switch to building fixture images with `docker buildx bake` instead of pulling published ones, and `docker/docker-bake.hcl`'s backend list drops `gemini` and adds `agy` accordingly.
- **`README.md`**, **`docs/design/sandbox-auth-projection.md`**, **`skills/helix/references/{run-debug,toml}.md`** — every backend list, example, and the ADR's readiness table updated. The ADR is explicit that `agy`'s manifest is derived-but-not-yet-confirmed-against-a-real-grant.
- **Tests** — `tests/unit/{test_sandbox,test_mutator,test_config,test_budget}.py` and `tests/integration/test_sandbox_auth_isolation_docker.py`: gemini-specific tests removed or replaced with agy-shaped ones (including a live-Docker test proving the tmpfs pre-owning behavior end to end); generic per-backend parametrizations pick up `agy` automatically since they're driven off `AUTH_CREDENTIAL_MANIFEST`/`BACKENDS` rather than a hardcoded list.
- **`CHANGELOG.md`** — new `[Unreleased]` **BREAKING** entry.

## Open unknowns (stated honestly, not glossed over)

1. **The credential file's exact JSON shape.** The path is confirmed by `strace` against a real binary. A synthetic OAuth-token-shaped blob at that path makes the CLI open it and then fail with `unknown auth method:` rather than `not logged in` — meaning the real file likely carries (or is paired with a settings field naming) an explicit auth-method discriminator, similar to what `gemini`'s removed `AUTH_SYNTHESIZED_SIBLINGS` entry handled. `AUTH_SYNTHESIZED_SIBLINGS` deliberately has **no** `agy` entry — HELIX's seed mechanism only ever `cp`s the credential file opaquely, so the manifest works without knowing the shape. The first real `helix sandbox login agy` inside the runner image will confirm it and reveal whether a sibling is needed.
2. **Whether `agy --print` takes the prompt as a trailing positional argument.** Every other HELIX backend does; `agy --help` doesn't show the exact grammar, and confirming it would require an actual `--print` call, which is out of scope for this migration (no spend, no generation — see the hard limits this work operated under). `_build_backend_args`'s `agy` branch follows the established convention with a comment noting this is confirmed on first real use.

## Explicitly out of scope

- The 0.2.1 → 0.3.0 version bump (see above — must land with it, not raced against it).
- `--json-schema` / enforcing the `HELIX_RESULT` contract at the CLI level — a different subsystem (the evaluator/prompt contract, not backend registration), and a real follow-up.
- A dedicated `agy` transcript tool-event counter — needs a real `--output-format json` transcript sample to write correctly; falls back to the generic parser until then.
- The `evolve --backend` hardcoded-list-vs-`BACKENDS` duplication in `cli.py` is a real, unrelated drive-by cleanup opportunity, deliberately not bundled in here.

## Test plan

- [x] `uv run python -m pytest` — 982 passed, 0 skipped (all 5 previously-skipped `agy` `docker_integration` cases now run for real: CI builds the `agy` fixture image itself instead of needing it published)
- [x] `HELIX_DOCKER_TESTS_STRICT=1 uv run python -m pytest tests/integration/ -m docker_integration -q -rs` — 27 passed, 0 skipped (one fewer than the 28-test baseline: `gemini`'s dedicated auth-method-gate test was removed with no direct replacement, since exercising it live would require an `agy --print` call, which the hard limits this work operated under forbid even against a synthetic credential)
- [x] `uv run ruff check src/ tests/` — clean
- [x] `uv run mypy --strict src/helix/` — clean
- [x] `docker/agy.Dockerfile` actually built (`helix-runner-base` → `helix-runner-agy`) and `agy --version` confirmed working both as root and as the real uid-1000 runtime user
- [x] Confirmed the five real `helix-auth-*` credential volumes on the build host were untouched throughout (name count only, never read)
- [x] Real CI `docker-integration`, rebased onto #61's build-based job: green, 982/0 as above; fixture build ~2m14s cold (new `agy` cache scope) after #61's already-warm scopes stayed cheap
- Diff: +2859/-393 lines across 24 files (includes #61's commits, expected for a stacked branch)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,9 +22,16 @@ jobs:
   docker-integration:
     # Real-Docker proof that per-candidate auth volumes actually isolate
     # credentials. Every credential used is synthetic, so this job needs no
-    # backend secrets: the runner images are anonymously pullable from GHCR,
-    # and the seeding helper runs in the backend's own runner image rather
-    # than introducing a third-party one.
+    # backend secrets.
+    #
+    # The fixture images are built from this checkout's own Dockerfiles
+    # rather than pulled from GHCR: a pulled image only reflects the last
+    # *published* build, so a PR that edits a Dockerfile would be tested
+    # against code it isn't changing, and a PR that adds a new backend could
+    # never go green -- its image can't exist on GHCR until after it merges.
+    # Building locally means every PR is validated against exactly the
+    # images its own diff produces. The seeding helper still runs in the
+    # backend's own runner image rather than introducing a third-party one.
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
@@ -32,13 +39,23 @@ jobs:
         with:
           python-version: "3.12"
       - run: uv sync --all-extras
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
       # The tests skip when a fixture image is absent. A fresh runner has
-      # none, so without this pull the job would pass while running nothing.
-      - name: Pull fixture images
-        run: |
-          for b in claude codex cursor gemini opencode; do
-            docker pull -q "ghcr.io/ke7/helix-evo-runner-$b:latest"
-          done
+      # none, so without this build the job would pass while running
+      # nothing. `docker buildx bake` (see docker/docker-bake.hcl) builds
+      # the base image once, then every backend's Dockerfile against it in
+      # parallel, using Buildx's named-context feature to point each
+      # backend's `FROM` straight at the base build's own output instead of
+      # a registry -- so build order only matters between the base and the
+      # backends, never among the backends themselves. Each target loads
+      # into the local Docker engine (`--load`, no registry push) under the
+      # exact tag `DEFAULT_BACKEND_IMAGES` in `src/helix/backends.py`
+      # resolves, and each caches independently through the GitHub Actions
+      # cache so an unrelated change doesn't invalidate every other image's
+      # layers.
+      - name: Build fixture images
+        run: docker buildx bake -f docker/docker-bake.hcl --load
       # STRICT=1 turns any remaining skip into a failure, so this job can
       # never be green without actually exercising the isolation tests.
       - run: uv run pytest tests/integration/ -m docker_integration -q -rs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,4 +31,16 @@ jobs:
         with:
           python-version: "3.12"
       - run: uv sync --all-extras
-      - run: uv run pytest tests/integration/ -m docker_integration -q
+      # The tests skip when a fixture image is absent. A fresh runner has
+      # none, so without this pull the job would pass while running nothing.
+      - name: Pull fixture images
+        run: |
+          docker pull -q python:3.13-alpine
+          for b in claude codex cursor gemini opencode; do
+            docker pull -q "ghcr.io/ke7/helix-evo-runner-$b:latest"
+          done
+      # STRICT=1 turns any remaining skip into a failure, so this job can
+      # never be green without actually exercising the isolation tests.
+      - run: uv run pytest tests/integration/ -m docker_integration -q -rs
+        env:
+          HELIX_DOCKER_TESTS_STRICT: "1"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
     # Real-Docker proof that per-candidate auth volumes actually isolate
     # credentials. Every credential used is synthetic, so this job needs no
     # backend secrets; the runner images are anonymously pullable from GHCR
-    # and the seeding helper runs in python:3.13-alpine.
+    # and the seeding helper runs in the agent's own backend runner image.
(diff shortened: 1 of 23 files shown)


KE7:
Folded into #61.

Both changes touch the same surface — `AUTH_CREDENTIAL_MANIFEST`, `AUTH_MOUNT_DESTINATIONS` and the backend registry — so keeping them apart meant the stack added a gemini credential manifest in one commit and deleted it in the next, with the design doc briefly recording gemini as ready for a backend that was already being removed.

`impl/auth-two-modes` now carries this commit directly (`d12e2cb..77705b6`, fast-forward, nothing rewritten). The combined branch has the final backend set — agy, claude, codex, cursor, opencode — and gemini never appears as a selectable backend at any point in it.

No content is lost by closing this: the commit is unchanged and now lives on #61.

